PR #13, opened 2019-12-18
[fbeV1] Publishing COM computed from estimated base state

prashanthr05:
This PR does the following,
- computes COM state from estimated base pose, base velocity and kinematic measurements
The COM state is available on the port
`/` estimator-prefix `/center_of_mass/state:o` in serialziation `x  y z vx vy vz` 
- fixes deprecation of yarp::os::Mutex and yarp::os::LockGuard by using std::mutex and std::lock_guard

cc @S-Dafarra

### DIFF
--- a/devices/baseEstimatorV1/include/baseEstimatorV1.h
+++ b/devices/baseEstimatorV1/include/baseEstimatorV1.h
@@ -394,7 +394,19 @@ namespace yarp {
          */
         bool updateBaseVelocityWithIMU();
 
+        /**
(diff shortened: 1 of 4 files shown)


prashanthr05:
Merged, thanks!

prashanthr05:
I have tested it on simulation. The `/base-estimator/center_of_mass/state:o` port appears and streams estimated com state as soon as a `startFloatingBaseEstimator` command is sent on the RPC port  through `yarp rpc /base-estimator/rpc`

PR #15, opened 2020-02-01
Merge devel in master

traversaro:
As master is empty, it should  not be a big problem. : )


traversaro:
@prashanthr05 can we proceed on this? I think it should not impact anyone working with this repo.

prashanthr05:
For me, it's a green flag.

As an exhaustive user of this estimator, I would like to have @GiulioRomualdi opinion on this as well.
Do you think we can merge into master or would you like something specific to be addressed before merging into master?

traversaro:
@prashanthr05 as you are the mantainer feel free to merge, thanks!

prashanthr05:
Merged, thanks!

PR #16, opened 2020-02-01
Migrate wholebodydynamics and related devices and libraries in whole-body-estimators

traversaro:
I originally planned to migrate also the history of the code, but as I was not expert in the procedure, I realized that the effort was probably not worth the gain. 

If anyone is interested, will be found in
https://github.com/robotology-legacy/codyco-modules (at the moment the link is not active as the repo still needs to be migrated). 

There are several parts of the code that would need to be cleaned up/removed deprecations/etc/etc, but I preferred to reduce at the minimum the modifications to the code in codyco-modules. Once the code has been merged, we can do the all the cleanup, in a clear and trackable way. 

cc @fjandrad @HosameldinMohamed unfortunatly any open PR needs to be re-created against the new repo, homever it should not be difficult as the files location is similar.


devices/CMakeLists.txt
```diff
@@ -3,4 +3,7 @@
 # GNU Lesser General Public License v2.1 or any later version.
 
 add_subdirectory(baseEstimatorV1)
-
+add_subdirectory(wholeBodyDynamics)
+add_subdirectory(virtualAnalogClient)
```
prashanthr05: I suppose these devices were added to YARP during these years. Could we add a small note or open an issue as to refactor out these devices out of this repository? And add the necessary documentation to use the YARP devices instead.

devices/wholeBodyDynamics/GravityCompensationHelpers.cpp
```diff
@@ -0,0 +1,222 @@
+#include "GravityCompensationHelpers.h"
```
prashanthr05: It must be noted that this class was ported to iDynTree https://github.com/robotology/idyntree/blob/master/src/estimation/include/iDynTree/Estimation/GravityCompensationHelpers.h

Might be useful to refactor out of this repository, in the future.

idl/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_subdirectory(wholeBodyDynamicsSettings)
+add_subdirectory(wholeBodyDynamics_IDLServer)
+add_subdirectory(floatingBaseEstimatorRPC)
```
prashanthr05: Do you think this `floatingBaseEstimatorRPC` might cause confusion going along with`baseEstimatorV1`?

idl/floatingBaseEstimatorRPC/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+# Copyright (C) 2016 Istituto Italiano di Tecnologia  iCub Facility
+# Authors: Silvio Traversaro
+# CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+
+yarp_idl_to_dir(codyco/floatingBaseEstimatorRPC.thrift ${CMAKE_CURRENT_SOURCE_DIR}/autogenerated WBDIDLSERVER_SRC WBDIDLSERVER_HEADERS WBDIDLSERVER_INCLUDES)
```
prashanthr05: Are we sure this will not cause any problems?

idl/floatingBaseEstimatorRPC/autogenerated/codyco_floatingBaseEstimatorRPC_thrift.cmake
```diff
@@ -0,0 +1,11 @@
...
+# All rights reserved.
+#
+# This software may be modified and distributed under the terms of the
+# BSD-3-Clause license. See the accompanying LICENSE file for details.
+
+## This is an automatically-generated file.
+## It could get re-generated if the ALLOW_IDL_GENERATION flag is on
+
+set(headers include/codyco/HomTransform.h;include/codyco/floatingBaseEstimatorRPC.h)
+set(sources src/HomTransform.cpp;src/floatingBaseEstimatorRPC.cpp)
```
prashanthr05: Is it safe to also copy autogenerated files?

libraries/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(ctrlLibRT)
```
prashanthr05: The purpose of this library was to enable the use of low pass filters on the encoder measurements.
I think these filters were ported to icub-main which is a dependency for this repository.
This must be kept in mind to refactor out this library.

devices/CMakeLists.txt
traversaro: Unfortunately I never actually migrated this devices in YARP.

prashanthr05: The device `genericSensorClient` though? Is it negative also for that?

devices/wholeBodyDynamics/GravityCompensationHelpers.cpp
traversaro: I think we can open issues on this, but I would avoid to mix improvements to the code to just the PR that move the code.

devices/CMakeLists.txt
traversaro: Unfortunately that as well.

devices/wholeBodyDynamics/GravityCompensationHelpers.cpp
prashanthr05: Definitely! I agree with that. It's just I noted these details which might be useful if collected together, for the future.

devices/CMakeLists.txt
prashanthr05: Ah, I see. My bad, then.

idl/CMakeLists.txt
traversaro: Good point, I need to check.

idl/floatingBaseEstimatorRPC/autogenerated/codyco_floatingBaseEstimatorRPC_thrift.cmake
traversaro: Again, they were committed also in the old repo, and they worked  fine. If we want to improve this, let's do this in a following PR.

libraries/CMakeLists.txt
traversaro: >  I think these filters were ported to icub-main which is a dependency for this repository.

Unfortunately, these filters were forked out of icub-main to remove the parts of the interface that prevented an efficient implementation (mainly the use of `yarp::sig::Vector`) so they cannot be directly replaced by their icub-main corresponding class. However, this is again something that should be tracked in an issue.

idl/floatingBaseEstimatorRPC/CMakeLists.txt
traversaro: See https://github.com/robotology/whole-body-estimators/pull/16#discussion_r380174507 .

libraries/CMakeLists.txt
prashanthr05: I didn't know that. Thank you for the information.

prashanthr05:
> It is crucial to test the migrated wholebodydynamics devices on the robot (iCubGenova04, active users of the device) with Yoga++ controller and walking controller.

This was done.

fjandrad:
Shall we press the Merge button @prashanthr05 @traversaro

traversaro:
Yes, I guess that that @prashanthr05 can do that.

prashanthr05:
> https://github.com/robotology/whole-body-estimators/pull/16/files#diff-0e39020ad0dd65979c2316050995dfa1 is a kate-swp files that needs to be removed.

This is still pending.

prashanthr05:
Let me delete that and push a commit. Then I will merge.

idl/CMakeLists.txt
prashanthr05: I think there will be no problem, because the `baseEstimatorV1` related RPC service is named as `floatingBaseEstimationRPC`, meanwhile here it is named `floatingBaseEstimatorRPC`. 
I guess this should be harmless.

prashanthr05: We can eventually remove this if not being used by other users.

prashanthr05:
@traversaro  @fjandrad  Merged into devel, Thank you!

prashanthr05:
@traversaro shall I go ahead and remove the `add-wbd` branch?

traversaro:
> @traversaro shall I go ahead and remove the `add-wbd` branch?

Yes thanks!

fjandrad:
So now, we wait for the inclusion of whole-body-estimators into robotology-superbuild should I open an issue for that? @traversaro

traversaro:
> So now, we wait for the inclusion of whole-body-estimators into robotology-superbuild should I open an issue for that? @traversaro

Yes, please open an issue for that, I plan to work on this  on the next days.

Issue #43, opened 2020-03-11
Merge devel into master and make a new release that includes wholebodydynamics and related devices


prashanthr05:
We just recently released a minor version of whole-body-estimators that includes the `baseEstimatorV1` device. We can proceed with another minor version release that includes the `wholebodydynamics` and related devices migrated from `codyco-modules`. This will enable an easy inclusion of whole-body-estimators in the [robotology/robotology-superbuild](https://github.com/robotology/robotology-superbuild) with correct version number.

HosameldinMohamed:
As per the T2T meeting with @traversaro, we shall do the following steps:

1. Bump the version in [CMakeLists.txt](https://github.com/robotology/whole-body-estimators/blob/devel/CMakeLists.txt) file.
2. Add a description in the [CHANGELOG](https://github.com/robotology/whole-body-estimators/blob/devel/CHANGELOG.md) file.
3. Merge the `devel` branch into the `master` branch (Adds `wholeBodyDynamics` and related devices to `master`).

CC @prashanthr05 @fjandrad

prashanthr05:
@traversaro  just a clarification. Should the points 1 and 2 indicated by @HosameldinMohamed be done in `master` during the release and then forward-merged into `devel`? or is it also a safe practice to do the necessary changes in `devel` itself and then merge into `master` for a release?

HosameldinMohamed:
> @traversaro just a clarification. Should the points 1 and 2 indicated by @HosameldinMohamed be done in `master` during the release and then forward-merged into `devel`? or is it also a safe practice to do the necessary changes in `devel` itself and then merge into `master` for a release?

 I think in future, the CHANGELOG file can be modified when introducing the changes (when merging the PR from a branch/fork to `devel`).

HosameldinMohamed:
> Now I was thinking to update CHANGELOG and CMakeLists in master after merging.

Actually, now that I think about it, if I do a commit in the `master` after the merge I will have to "forward merge" the `master` into `devel`. So maybe it's better to do one commit in `devel` then merge it to master.

prashanthr05:
Yes, probably it would be cleaner and avoid too many circular commits.

traversaro:
> 
> @traversaro just a clarification. Should the points 1 and 2 indicated by @HosameldinMohamed be done in `master` during the release and then forward-merged into `devel`? or is it also a safe practice to do the necessary changes in `devel` itself and then merge into `master` for a release?

[...]

PR #44, opened 2020-03-11
Add wholeBodyDynamics and its related devices

HosameldinMohamed:
This PR adds to the `master` branch the following YARP devices:

- genericSensorClient.
- virtualAnalogClient.
- virtualAnalogRemapper.
- wholeBodyDynamics.

Related issue #43


prashanthr05:
cc @HosameldinMohamed would you like to address this comment? https://github.com/robotology/whole-body-estimators/issues/43#issuecomment-597572408

HosameldinMohamed:
> cc @HosameldinMohamed would you like to address this comment? #43 (comment)

I am not sure if we should put `set VARS_PREFIX` as for example in `iDynTree`.

https://github.com/robotology/idyntree/blob/685c0803fda91e5f8046efcd55fba3bcdc9f6907/CMakeLists.txt#L24

Because we don't use `find_package(wholeBodyEstimators)`, right?

traversaro:
> 
> 
> > cc @HosameldinMohamed would you like to address this comment? #43 (comment)
> 
> I am not sure if we should put `set VARS_PREFIX` as for example in `iDynTree`.
> 
> https://github.com/robotology/idyntree/blob/685c0803fda91e5f8046efcd55fba3bcdc9f6907/CMakeLists.txt#L24
> 
> Because we don't use `find_package(wholeBodyEstimators)`, right?
> 
> CC @traversaro

Sorry, I am not sure what you are asking. iDynTree uses an old way of setting the version, the new one used by  this project (the version variable  in the project CMake function) is much better.

HosameldinMohamed:
> Sorry, I am not sure what you are asking. iDynTree uses an old way of setting the version, the new one used by this project (the version variable in the project CMake function) is much better.

Ah okay, thanks, you just clarified it. I thought the different way of setting the version has a different result.

PR #45, opened 2020-03-11
Add support for compensating temperature in Force/Torque sensors readings and for specify user offline offset 

fjandrad:
This PR ports the functionality of considering temperature coefficients in the FT sensors and allows the user to enable if he wants to use an FT offset estimated offline.

It was orignally in [codyco-modules](https://github.com/robotology/codyco-modules/pull/281) and now ported to this repo.

### DIFF
--- a/devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
+++ b/devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
@@ -68,15 +68,12 @@ const double& SixAxisForceTorqueMeasureProcessor::tempOffset() const
 
 iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::filt(const iDynTree::Wrench& input) const
 {
-    Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input)-toEigen(m_offset);
-    iDynTree::Wrench ret;
-    fromEigen(ret,retEig);
-    return ret;
+    return filt(input, 0.0);
 }
 
 iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::filt(const iDynTree::Wrench & input, const double & temperature) const
 {
-    Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input)-toEigen(m_offset)+toEigen(m_temperatureCoefficients)*(temperature-m_tempOffset);    
+    Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input)-toEigen(m_offset)+toEigen(m_temperatureCoefficients)*(temperature-m_tempOffset);
     iDynTree::Wrench ret;
     fromEigen(ret,retEig);
 
@@ -86,12 +83,7 @@ iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::filt(const iDynTree::Wrench
 
 iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::applySecondaryCalibrationMatrix(const iDynTree::Wrench& input) const
 {
-    Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input);
-
-    iDynTree::Wrench ret;
-    fromEigen(ret,retEig);
-
-    return ret;
+    return applySecondaryCalibrationMatrix( input, 0.0);
 }
 
 iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::applySecondaryCalibrationMatrix(const iDynTree::Wrench& input, const double & temperature) const
@@ -105,5 +97,3 @@ iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::applySecondaryCalibrationMa
 }
 
 }
-
-
--- a/devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
+++ b/devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
@@ -1,6 +1,7 @@
 #include "SixAxisForceTorqueMeasureHelpers.h"
 #include <iDynTree/Core/EigenHelpers.h>
 
+
 namespace wholeBodyDynamics
 {
 
@@ -9,6 +10,9 @@ SixAxisForceTorqueMeasureProcessor::SixAxisForceTorqueMeasureProcessor()
     // Initialize the affice function to be the identity
     toEigen(m_secondaryCalibrationMatrix).setIdentity();
     m_offset.zero();
+    toEigen(m_temperatureCoefficients).setZero();
+    m_tempOffset=0.0;
+    m_estimated_offset.zero();
 }
 
 iDynTree::Matrix6x6& SixAxisForceTorqueMeasureProcessor::secondaryCalibrationMatrix()
@@ -16,10 +20,20 @@ iDynTree::Matrix6x6& SixAxisForceTorqueMeasureProcessor::secondaryCalibrationMat
     return m_secondaryCalibrationMatrix;
 }
 
+iDynTree::Vector6& SixAxisForceTorqueMeasureProcessor::temperatureCoefficients()
+{
+    return m_temperatureCoefficients;
+}
+
 
 const iDynTree::Matrix6x6& SixAxisForceTorqueMeasureProcessor::secondaryCalibrationMatrix() const
 {
     return m_secondaryCalibrationMatrix;
+}
+
+const iDynTree::Vector6& SixAxisForceTorqueMeasureProcessor::temperatureCoefficients() const
+{
+    return m_temperatureCoefficients;
 }
 
 iDynTree::Wrench& SixAxisForceTorqueMeasureProcessor::offset()
@@ -32,12 +46,40 @@ const iDynTree::Wrench& SixAxisForceTorqueMeasureProcessor::offset() const
     return m_offset;
 }
 
+iDynTree::Wrench& SixAxisForceTorqueMeasureProcessor::estimatedOffset()
+{
+    return m_estimated_offset;
+}
+
+const iDynTree::Wrench& SixAxisForceTorqueMeasureProcessor::estimatedOffset() const
+{
+    return m_estimated_offset;
+}
+
+double& SixAxisForceTorqueMeasureProcessor::tempOffset()
+{
+    return m_tempOffset;
+}
+
+const double& SixAxisForceTorqueMeasureProcessor::tempOffset() const
+{
+    return m_tempOffset;
+}
+
 iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::filt(const iDynTree::Wrench& input) const
 {
     Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input)-toEigen(m_offset);
-
     iDynTree::Wrench ret;
     fromEigen(ret,retEig);
+    return ret;
+}
+
+iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::filt(const iDynTree::Wrench & input, const double & temperature) const
+{
+    Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input)-toEigen(m_offset)+toEigen(m_temperatureCoefficients)*(temperature-m_tempOffset);    
+    iDynTree::Wrench ret;
+    fromEigen(ret,retEig);
+
 
     return ret;
 }
@@ -45,6 +87,16 @@ iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::filt(const iDynTree::Wrench
 iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::applySecondaryCalibrationMatrix(const iDynTree::Wrench& input) const
 {
     Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input);
+
+    iDynTree::Wrench ret;
+    fromEigen(ret,retEig);
+
+    return ret;
+}
+
+iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::applySecondaryCalibrationMatrix(const iDynTree::Wrench& input, const double & temperature) const
+{
+    Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input)+toEigen(m_temperatureCoefficients)*(temperature-m_tempOffset);
 
     iDynTree::Wrench ret;
     fromEigen(ret,retEig);
--- a/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
+++ b/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
@@ -11,7 +11,7 @@
 
 #include <iDynTree/yarp/YARPConversions.h>
 #include <iDynTree/Core/Utils.h>
-
+#include <iDynTree/Core/EigenHelpers.h>
 #include <cassert>
 #include <cmath>
 
@@ -43,6 +43,10 @@ WholeBodyDynamicsDevice::WholeBodyDynamicsDevice(): RateThread(10),
     calibrationBuffers.measurementSumBuffer.resize(0);
     calibrationBuffers.nrOfSamplesToUseForCalibration = 0;
     calibrationBuffers.nrOfSamplesUsedUntilNowForCalibration = 0;
+    // temperature measuremnts vector
+    tempMeasurements.resize(0);
+    ftTempMapping.resize(0);
+    prevFTTempTimeStamp=0;
 
 }
 
@@ -142,46 +146,34 @@ void addVectorOfStringToProperty(yarp::os::Property& prop, std::string key, std:
     return;
 }
 
-bool getUsedDOFsList(os::Searchable& config, std::vector<std::string> & usedDOFs)
+bool getConfigParamsAsList(os::Searchable& config,std::string propertyName , std::vector<std::string> & list)
 {
     yarp::os::Property prop;
     prop.fromString(config.toString().c_str());
-
-    yarp::os::Bottle *propAxesNames=prop.find("axesNames").asList();
-    if(propAxesNames==0)
+    yarp::os::Bottle *propNames=prop.find(propertyName).asList();
+    if(propNames==nullptr)
     {
-       yError() <<"WholeBodyDynamicsDevice: Error parsing parameters: \"axesNames\" should be followed by a list\n";
-       return false;
+        yError() <<"WholeBodyDynamicsDevice: Error parsing parameters: \" "<<propertyName<<" \" should be followed by a list\n";
+        return false;
     }
 
-    usedDOFs.resize(propAxesNames->size());
-    for(int ax=0; ax < propAxesNames->size(); ax++)
+    list.resize(propNames->size());
+    for(auto elem=0u; elem < propNames->size(); elem++)
     {
-        usedDOFs[ax] = propAxesNames->get(ax).asString().c_str();
+        list[elem] = propNames->get(elem).asString().c_str();
     }
 
     return true;
 }
 
+bool getUsedDOFsList(os::Searchable& config, std::vector<std::string> & usedDOFs)
+{
+    return getConfigParamsAsList(config,"axesNames",usedDOFs);
+}
+
 bool getGravityCompensationDOFsList(os::Searchable& config, std::vector<std::string> & gravityCompensationDOFs)
 {
-    yarp::os::Property prop;
-    prop.fromString(config.toString().c_str());
-
-    yarp::os::Bottle *propAxesNames=prop.find("gravityCompensationAxesNames").asList();
-    if(propAxesNames==0)
-    {
-       yError() <<"wholeBodyDynamics : Error parsing parameters: \"gravityCompensationAxesNames\" should be followed by a list\n";
-       return false;
-    }
-
-    gravityCompensationDOFs.resize(propAxesNames->size());
-    for(int ax=0; ax < propAxesNames->size(); ax++)
-    {
-        gravityCompensationDOFs[ax] = propAxesNames->get(ax).asString().c_str();
-    }
-
-    return true;
+    return getConfigParamsAsList(config,"gravityCompensationAxesNames",gravityCompensationDOFs);
 }
 
 
@@ -707,6 +699,54 @@ bool WholeBodyDynamicsDevice::openExternalWrenchesPorts(os::Searchable& config)
     return ok;
 }
 
+bool WholeBodyDynamicsDevice::openMultipleAnalogSensorRemapper(os::Searchable &config)
+{
+    // Pass to the remapper just the relevant parameters (sensorList)
+    yarp::os::Property prop;
+    prop.fromString(config.toString().c_str());
+    yarp::os::Property propMASRemapper;
+    yarp::os::Bottle & propMasNames=prop.findGroup("multipleAnalogSensorsNames");
+    propMASRemapper.put("device","multipleanalogsensorsremapper");
+    bool ok=false;
+    for (auto types=1u;types<propMasNames.size();types++){
+        yarp::os::Bottle * mas_type = propMasNames.get(types).asList();
+        if( mas_type->size() != 2 || mas_type->get(1).asList() == nullptr )
+        {
+            yError() << "wholeBodyDynamics: multipleAnalogSensorsNames group is malformed (" << mas_type->toString() << "). ";
+            return false;
+        }
+        else {
+            ok=true;
+        }
+
+        propMASRemapper.addGroup(mas_type->get(0).asString());
+        yarp::os::Bottle  MASnames;
+        yarp::os::Bottle & MASnamesList= MASnames.addList();
+        for(auto i=0u; i < mas_type->get(1).asList()->size(); i++)
+        {
+            MASnamesList.addString(mas_type->get(1).asList()->get(i).asString());
+        }
+        propMASRemapper.put(mas_type->get(0).asString(),MASnames.get(0));
+
+    }
+    ok = multipleAnalogRemappedDevice.open(propMASRemapper);
+    if( !ok )
+    {
+        return ok;
+    }
+    // View relevant interfaces for the multipleAnalogRemappedDevice
+    ok = ok && multipleAnalogRemappedDevice.view(remappedMASInterfaces.temperatureSensors);
+    ok = ok && multipleAnalogRemappedDevice.view(remappedMASInterfaces.ftMultiSensors);
+    ok = ok && multipleAnalogRemappedDevice.view(remappedMASInterfaces.multwrap);
+    if( !ok )
+    {
+        yError() << "wholeBodyDynamics : openMultipleAnalogSensorRemapper : error while opening the necessary interfaces in multipleAnalogRemappedDevice";
+        return ok;
+    }
+
+   return true;
+}
+
 bool WholeBodyDynamicsDevice::openFilteredFTPorts(os::Searchable& config)
 {
     bool ok = true;
@@ -763,6 +803,10 @@ void WholeBodyDynamicsDevice::resizeBuffers()
     calibrationBuffers.predictedSensorMeasurementsForCalibration.resize(estimator.sensors());
 
     ftProcessors.resize(nrOfFTSensors);
+
+    // Resize temperature sensor inside F/T sensors
+    this->tempMeasurements.resize(nrOfFTSensors,0.0);
+    this->ftTempMapping.resize(nrOfFTSensors,-1);
 
     // Resize filters
     filters.init(nrOfFTSensors,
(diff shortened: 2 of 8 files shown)


devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -522,6 +514,54 @@ bool WholeBodyDynamicsDevice::openExternalWrenchesPorts(os::Searchable& config)
...
+    }
+    ok = multipleAnalogRemappedDevice.open(propMASRemapper);
+    if( !ok )
+    {
+        return ok;
+    }
+    // View relevant interfaces for the multipleAnalogRemappedDevice
+    ok = ok && multipleAnalogRemappedDevice.view(remappedMASInterfaces.temperatureSensors);
+    ok = ok && multipleAnalogRemappedDevice.view(remappedMASInterfaces.ftMultiSensors);
+    ok = ok && multipleAnalogRemappedDevice.view(remappedMASInterfaces.multwrap);
```
prashanthr05: Could you also explain why we use a IMultipleWrapper interface here? Since, we already have access to the MASInterfaces through the MASRemapper?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -2083,6 +2377,30 @@ bool WholeBodyDynamicsDevice::resetOffset(const std::string& calib_code)
...
+bool WholeBodyDynamicsDevice::usePreEstimatedOffset()
+{
+    yarp::os::LockGuard guard(this->deviceMutex);
+
+    yWarning() << "wholeBodyDynamics : using offset estimated offline.";
+
+    for(size_t ft = 0; ft < this->getNrOfFTSensors(); ft++)
+    {
+        double norm = iDynTree::toEigen(ftProcessors[ft].estimatedOffset()).norm();
+        if (norm!= 0.0){ // if norm is exactly 0.0 it means there was no offset  in the configuration file so it should be skipped
```
prashanthr05: this seems dangerous ;D This needs to be documented somewhere visible to the user.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1161,14 +1425,39 @@ void convertVectorFromDegreesToRadians(iDynTree::VectorDynSize & vector)
...
+    bool TempSensorReadCorrectly = true;
+    double timeFTStamp=yarp::os::Time::now();
+    bool readTemperatureSensorThisTime=false;
+    yarp::dev::MAS_status    sensorStatus;
     for(size_t ft=0; ft < estimator.sensors().getNrOfSensors(iDynTree::SIX_AXIS_FORCE_TORQUE); ft++ )
     {
         iDynTree::Wrench bufWrench;
         int ftRetVal = ftSensors[ft]->read(ftMeasurement);
+//TODO: change 0.55 for a variable. 0.55 was related to the fact that temperature sensors are updated once every second
+        if (timeFTStamp-prevFTTempTimeStamp>0.55){
```
prashanthr05: magic number! Please address this TODO in this PR.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1029,6 +1239,60 @@ bool WholeBodyDynamicsDevice::attachAllFTs(const PolyDriverList& p)
...
+    // Check if the MASremapper and the estimator have a consistent number of ft sensors
+    int tempSensors = 0;
+    tempSensors=static_cast<int>( remappedMASInterfaces.temperatureSensors->getNrOfTemperatureSensors());
+    if( tempSensors > static_cast<int>( estimator.sensors().getNrOfSensors(iDynTree::SIX_AXIS_FORCE_TORQUE)) )
+    {
+        yError() << "wholeBodyDynamics : The multipleAnalogRemappedDevice has more sensors than those in the open estimator ft sensors list";
+        return false;
+    }
+
+    // Temporary code to check name of sensors, assuming the temperature sensor name will be the same as the ft sensor
```
prashanthr05: Is this temporary code necessary or, can it be removed?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1029,6 +1239,60 @@ bool WholeBodyDynamicsDevice::attachAllFTs(const PolyDriverList& p)
...
+    int checkCounter=0;
+    std::string ftName;
+    std::string tempName;
+    int ftMap=-1;
+    //for (auto & sensorName : tempDeviceNames){
```
prashanthr05: Could you remove this stranded line?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
fjandrad: Added in the description of the function in the .h file

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
fjandrad: Well is what was agreed when developing the logic it needs to set the proper names in the configuration files of the sensor. So it is not really temporary anymore, removed this line.
The problem is I do not know where to document this at the moment.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
fjandrad: I was trying to prepare for moving to newer C++ standard but yeah it doesn't make sense to be there for nothing. Deleting

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
fjandrad: I will address it then. I guess it will be another configuration parameter.

devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
```diff
@@ -1,6 +1,7 @@
 #include "SixAxisForceTorqueMeasureHelpers.h"
 #include <iDynTree/Core/EigenHelpers.h>
 
+
```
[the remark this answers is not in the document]
fjandrad: done

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
fjandrad: Not sure I got the question or know the answer since I do not remember much how they work at the moment need to reread some of it

fjandrad:
> The relevant changes in the configuration files are missing in this PR?

It is perfectly back-compatible so no change is required in the configuration file. What needs to be done is to add the configuration file with the required variables in robot configuration and then we may put a link in the PR. what do you think @prashanthr05

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
prashanthr05: I had the same thoughts :D that's why I wanted to understand.

prashanthr05:
> What needs to be done is to add the configuration file with the required variables in robot configuration and then we may put a link in the PR

Agreed. 

But should we also add such a configuration file in the `app` folder of the `wholebodydynamics` device in this repository?

fjandrad:
> But should we also add such a configuration file in the app folder of the wholebodydynamics device in this repository?

I do not know what is the best policy here. As a possible user of these features I would say it is nice to have a default or dummy with already the variables added so that I can modify them later on. It could also make other users aware that those functionalities exist by looking at the parameters there.
But it could also make noise since they are not strictly required parameters and a new user might think they are because they are there. @traversaro what do you think?

traversaro:
> 
> 
> > What needs to be done is to add the configuration file with the required variables in robot configuration and then we may put a link in the PR
> 
> Agreed.
> 
> But should we also add such a configuration file in the `app` folder of the `wholebodydynamics` device in this repository?

I think we should start to deprecate the configuration files in the `app` folder, and just leave there the one for simulated robots. If we keep the configuration files in robots-configuration, then we should avoid that the same files are duplicated here. So I think we can avoid updating the files in `app` .

fjandrad:
# Testing this PR
The original testing idea was to compare if it was useful to add the temperature and in that case if we could use the offset estimated with it. The conclusion of such experiments was that yes we benefit from the temperature and more over it enables to use an offset estimated offline since main drift cause ( the temperature ) is taken care of by the temperature coefficients.  Results were reported in [Six-Axis Force Torque Sensor Model-Based In Situ Calibration Method and Its Impact in Floating-Based Robot Dynamic Performance](https://www.mdpi.com/1424-8220/19/24/5521).

So to test this PR what we would need to do is insert the temperature coefficients and offset candidate in the configuration file.

 An example of combination of uses enabled by this PR are here 
[temperature_conf_files.zip](https://github.com/robotology/whole-body-estimators/files/4355313/temperature_conf_files.zip).  They are prepared to launch an external whole-body-dynamics device for testing without risking the actual robot. 

:warning:  if we compare the offset recently estimated using calib all vs the pre-estimated offset we might see a small benefit from calib all in comparison to the offline offset. On the long run, the estimated offset will keep its good level of performance while the performance of the offset estimated with calib all can drop by a big amount. So the use of this feature has two main scenarios in which it has a good advantage: 
1. When immediatly starting the robot, since it warms up fast the first time changing the offset 'fast' 
1. When we schedule tests that will last a long time or plan to use the robot for extended periods of time ( more than 1 hour).



Once validated, we should replace the current wholebodydynamics.xml by one with the variables required for the temperature compensation with offline estimated offsets. The changes will be inserted in robots-configuration (actual file still pending)


## Testing scenarios
The idea is that even when using the robot for a long period of time the offset does not need to be estimated again. 

### Reasonable values
1. launch yarprobotinterface with default setup
2. launch an external wholebodydynamics (wbd) device with the temperature coefficients and offset
1. for the externald wbd device use the rpc usePreEstimatedOffset function.
3. verify the values are reasonable in the external device by putting robot on the ground in two feet
4. verify the values are reasonable in single support ( there should be some default home positions in yarpmotorgui )

### Performance verification
1. Start the robot, when the robot has not been used in a long time and motors were off. ( best scenario is early morning ). 
1. Launch external wbd device with the temperature coefficients and offset
1. For the externald wbd device use the rpc usePreEstimatedOffset function.
1. Run a couple of experiments like Yoga or walking ( 2 yogas one for each foot is prefered ) while logging the experiment with yarp data dumper both the local and external wbd force information ( filtered ft port) and required information to estimate offline the expected wrenches ( joint positions, imu , ft sensor information ).
1. Wait **with the robot on** for 20 min then launch other set experiments.
1. Repeat previous step 4 or 5 times
1. Use iDyntree scripts to verify what was the expected value when standing on single support

It is expected to see that the one with temperature compensation has much better performance in th e later experiments than without compensation.

:exclamation: Is important to remember **not** to do calib all again at any time to have a clear difference in performance.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
fjandrad: In general, remember I copied the same style of controlboardremappers. At the time I was doing it, I understood why now I do not remember hehe

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
```diff
@@ -188,6 +189,50 @@ class wholeBodyDynamicsDeviceFilters
  *      </group>
  * \endcode
  *
+ * * \subsection TemperatureCoefficients
+ * This device support to specify a set of temperature coefficients matrix to apply on the top of the (already calibrated) measure coming from the F/T sensors.
+ * This feature is meant to be experimental, and will be removed at any time.
```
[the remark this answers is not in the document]
fjandrad: Done

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
fjandrad: created variable checkTemperatureEvery_seconds with default value of 0.55

prashanthr05:
@fjandrad have you already pushed your changes to the remote? because I do not see them. 
Also, I wonder why the CI is not starting in this PR?

fjandrad:
@prashanthr05 not yet got distracted with ISAAC stuff. No idea about the CI, its so clean that it believes it does not need to check :laughing:  :rofl:

traversaro:
>  Also, I wonder why the CI is not starting in this PR?

I guess because the PR is based on an old devel branch without CI. I suggest to rebase on the latest devel to get CI.

fjandrad:
It's starting now. Is it possible the PR was created before we added CI functionalities to the repo?
Note that I did not rebase for the CI to start.

fjandrad:
It stopped the others because Windows failed. Is there a way to allow it to check it anyway?

prashanthr05:
> It stopped the others because Windows failed. Is there a way to allow it to check it anyway?

According to [this answer](https://github.community/t5/GitHub-Actions/Why-a-matrix-step-will-be-canceled-if-another-one-failed/td-p/30920), this is due to a [fail-fast](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) behavior and by default it is set to true. I will open a PR.

traversaro:
> 
> 
> It stopped the others because Windows failed. Is there a way to allow it to check it anyway?

The Windows failure is something affecting several other repos, see https://github.com/dic-iit/bipedal-locomotion-controllers/issues/30 .

prashanthr05:
> > It stopped the others because Windows failed. Is there a way to allow it to check it anyway?
> 
> According to [this answer](https://github.community/t5/GitHub-Actions/Why-a-matrix-step-will-be-canceled-if-another-one-failed/td-p/30920), this is due to a [fail-fast](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) behavior and by default it is set to true. I will open a PR.

Relevant PR opened in https://github.com/robotology/whole-body-estimators/pull/52

fjandrad:
For testing this PR, a branch has been opened https://github.com/dic-iit/robots-configuration/tree/testTemperatureOffset . At the moment, it contains only the main wholebodydyinamcs.xml with the variables that gave best results for the paper. No files for comparison where directly added, but they are anyway contained [here](https://github.com/robotology/whole-body-estimators/pull/45#issuecomment-601255577).

devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
```diff
@@ -52,6 +94,16 @@ iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::applySecondaryCalibrationMa
     return ret;
 }
 
+iDynTree::Wrench SixAxisForceTorqueMeasureProcessor::applySecondaryCalibrationMatrix(const iDynTree::Wrench& input, const double & temperature) const
+{
+    Eigen::Matrix<double,6,1> retEig = toEigen(m_secondaryCalibrationMatrix)*toEigen(input)+toEigen(m_temperatureCoefficients)*(temperature-m_tempOffset);
```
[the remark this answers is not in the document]
fjandrad: Not sure if you meant this but...  the general math can be found [here](https://www.mdpi.com/1424-8220/19/24/5521/htm), Section 3.4.1 has some mentions of the secondary calibration matrix.

The math should be $C_s * r + C_t*(t-t_0)$

_$r$_ are the sensor measurements in N/Nm with the offset removed, _$C_s$_ the secondary calibration matrix, _$C_t$_ the emperature coefficients, _$t$_ the temperature and _$t_0$_ the temperature offset.

fjandrad:
I will leave this in ice-box until we can to the test in the lab with the real robot.  Next iteration hopefully.

fjandrad:
I tried testing the PR, sadly it failed:

- I first tried the rebased version of the PR, but the yarp of the computer and robot is not updated. So when building and installing I faced:
image

To avoid updating everything since I had no time I tried with the non rebased version.
I followed the steps in https://github.com/dic-iit/component_wholebody-teleoperation/issues/250#issue-578145346

Unfortunately is not working. I think it might be a problem on the names of the sensors configuration files, but unfortunately do not have time to verify it since I need to go.

The error that I could find in the logger is here:

image

A length problem just after reading the sensors. So after this line https://github.com/robotology/whole-body-estimators/blob/ed3bd4c1c06bd0afa885cc9b5c3bfbffd84c17d5/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L1241

I'm sorry I could not fix this before leaving. @prashanthr05 @traversaro If there is any doubt I can help with later let me know. I guess I wont be able to test this in the robot anymore.

prashanthr05:
I rebased the branch over the latest devel after fixing a couple of merge conflicts. We can schedule testing this PR in this iteration.

prashanthr05:
oops, instead of pressing comment, I pressed close and comment :D!

prashanthr05:
@HosameldinMohamed and I were testing this PR today.  It was failing in the `attachAllFTs()` phase.

We identified the problems to be the following,
- [Here](https://github.com/robotology/whole-body-estimators/pull/45/files#diff-12e87bffb77db5b00a60bde6f8de53d9R1568) we parse through the `ftDeviceNames` which includes both MAS FTs and analog FT names (size up to 10 for iCubGenova04), however, the for loop iterates over `ftList` which is a vector of analog sensors only (6 in number). So there is a mismatch in checking the sensor names between the added FT sensors and those in the URDF model.
-  a missing `return false` [here](https://github.com/robotology/whole-body-estimators/pull/45/files#diff-12e87bffb77db5b00a60bde6f8de53d9R1543) (however, this did not cause any problem, so far)

However, even fixing these did not allow us to successfully launch the wholebodydynamics device. This needs to be investigated further, possibly need to run a yarprobotinterface with valgrind to understand the root cause of failure, because this could not be interpreted from the logs. The yarprobotinterface just goes into a "freeze" state and then crashes after a while.

traversaro:
Is the  calibration working correctly or it is disabled?

prashanthr05:
> Is the calibration working correctly or it is disabled?

The calibration was running. We were able to see almost zero end-effector wrenches at the feet even without running a `calib all 300` through the rpc (not sure if we should be surprised about this, because we usually see a ~20 N offset on forces most times on the robot).

However, we still need to test the pre-estimated offset and temperature compensation. We couldn't fully understand the details in this [comment](https://github.com/robotology/whole-body-estimators/pull/45#issuecomment-601255577). Maybe @fjandrad is available sometime, we could have a quick call to discuss the procedure?

fjandrad:
```
We identified the problems to be the following,

  -   Here we parse through the ftDeviceNames which includes both MAS FTs and analog FT names (size up to 10 for iCubGenova04), however, the for loop iterates over ftList which is a vector of analog sensors only (6 in number). So there is a mismatch in checking the sensor names between the added FT sensors and those in the URDF model.
  -     a missing return false here (however, this did not cause any problem, so far)

```

This was on purpose since the MAS have both the interface for analog and MAS so when looking for analog sensors the MAS will also appear and are as of now the default mode for the FT sensors. The name in the overlapping analog sensors and MAS should be the same. But the name of the MAS is actually read from the sensors while for the analog they do not really matter. When testing my impression was that the name of the sensors in the hardware conf files might be not the ones expected of the urdf, but could not check further. I am available if you want to have a call.

prashanthr05:
> This was on purpose since the MAS have both the interface for analog and MAS so when looking for analog sensors the MAS sensors will also appear in the analog list and are as of now the default mode for the FT sensors. The name in the overlapping analog sensors and MAS should be the same

Eventhough they might be from the same device, we populate the vector of names by checking the interfaces which are different (ISixAxisForceTorqueSensors and IAnalogSensors). So in that case, it is also ok to maintain two separate vectors, because the compatibility check with the sensors present in the URDF model is done using the list populated by analog sensor interfaces.

traversaro:
> 
> 
> > This was on purpose since the MAS have both the interface for analog and MAS so when looking for analog sensors the MAS sensors will also appear in the analog list and are as of now the default mode for the FT sensors. The name in the overlapping analog sensors and MAS should be the same
> 
> Eventhough they might be from the same device, we populate the vector of names by checking the interfaces which are different (ISixAxisForceTorqueSensors and IAnalogSensors). So in that case, it is also ok to maintain two separate vectors, because the compatibility check with the sensors present in the URDF model is done using the list populated by analog sensor interfaces.

I think that is the case now, but if we could report this problem now will permit to fix them at the robot configuration level, so that in the future we will be able to migrate away from `IAnalogSensors` .

prashanthr05:
In that case, the MAS wrappers for the arm FT sensors are clearly missing here.

https://github.com/robotology/robots-configuration/tree/master/iCubGenova04/wrappers/FT

traversaro:
> and maybe boils down to the fact that the arm FT sensors are not strain2 sensors?

I am afraid you are right, all strain1 sensors uses the `embObjStrain` devices that at the moment does not support MAS interfaces.

fjandrad:
> I am afraid you are right, all strain1 sensors uses the embObjStrain devices that at the moment does not support MAS interfaces.

Having only strain 2 sensors would have been much easier also in the code.

traversaro:
> 
> 
> > I am afraid you are right, all strain1 sensors uses the embObjStrain devices that at the moment does not support MAS interfaces.
> 
> Having only strain 2 sensors would have been much easier also in the code.

Having only strain 2 sensors on all iCub in the world is a bit impossible for the few next years for logistical reasons, but making sure that all strain sensors also expose a MAS interface  is much easier, I will open appropriate issues for that.

HosameldinMohamed:
> > Is the calibration working correctly or it is disabled?

> 
> The calibration was running. We were able to see almost zero end-effector wrenches at the feet even without running a calib all 300 through the rpc (not sure if we should be surprised about this, because we usually see a ~20 N offset on forces most times on the robot).

@traversaro considering this line inside `attachAll` which is automatically called before `calib all`
https://github.com/robotology/whole-body-estimators/blob/772be87fc71798eb5ca4f3eb15dd4df49b1acf82/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L1387

should we disable also that while testing the `preEstimatedOffset`?

traversaro:
> should we disable also that while testing the preEstimatedOffset?

Exactly, if we want to use the `preEstimatedOffset` we don't want to compute them online via the calibration. Furthermore, you must not call `calib all` if you want to test with  `preEstimatedOffset`, otherwise the `calib all` will overwrite the offset that you specified with  `preEstimatedOffset`.

devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
```diff
@@ -9,19 +10,32 @@ SixAxisForceTorqueMeasureProcessor::SixAxisForceTorqueMeasureProcessor()
     // Initialize the affice function to be the identity
     toEigen(m_secondaryCalibrationMatrix).setIdentity();
     m_offset.zero();
+    toEigen(m_temperatureCoefficients).setZero();
+    m_tempOffset=0.0;
+    m_estimated_offset.zero();
```
fjandrad, line 15: @traversaro If you call calib all it should not overwrite the offset obtained from the configuration file since it is stored separately. At least that was the intention. This way it kept backward compatibility and the preestimated offset is not lost in case the user inadvertently used calib all.

@HosameldinMohamed  you could verify if its the case.

fjandrad: `m_estimated_offset` keeps the offset from the configuration file while the computed offset by calib all is always stored in `m_offset`. So calib all will overwrite temporarily the offest from the configuration file but it can be restored by calling the `usePreEstimated` offset function from rpc

traversaro: Thanks @fjandrad, I missed that detail.

fjandrad: Not entirely useless. You can alternate between one and the other calling `calib all `and `usePredefinedOffset` from rpc at any time.

traversaro: Yes, totally. I was actually referring to the `calib all` that from this comment https://github.com/robotology/whole-body-estimators/pull/45#issuecomment-659975787 I understood that @HosameldinMohamed is using for his tests.

HosameldinMohamed:
@prashanthr05 and I tested the first part of the https://github.com/robotology/whole-body-estimators/pull/45#issuecomment-601255577 launching an external device using the `wholebodydynamics-temp-offset.xml` config file attached in the comment. In addition to the usual device launched in iCub-head.

- The `FT_OFFSET`, `FT_TEMPERATURE_COEFFICIENTS` And `FT_SECONDARY_CALIBRATION` parameters are the same in the config file mentioned above.

- The performed activities are:

> ### Reasonable values
> 1. launch yarprobotinterface with default setup
> 2. launch an external wholebodydynamics (wbd) device with the temperature coefficients and offset
> 1. for the externald wbd device use the rpc usePreEstimatedOffset function.
> 3. verify the values are reasonable in the external device by putting robot on the ground in two feet
> 4. verify the values are reasonable in single support ( there should be some default home positions in yarpmotorgui )

- The robot was ON for a long time before recording the data (>1 hour).

- For `standing on two feet`, the dumped ports are:

- Left leg FT sensor:
left_leg_FT_force_X
left_leg_FT_force_Y
left_leg_FT_force_Z
left_leg_FT_torque_X
left_leg_FT_torque_Y
left_leg_FT_torque_Z



- Left leg End effector wrench:
left_leg_EndEffector_force_X
left_leg_EndEffector_force_Y
left_leg_EndEffector_force_Z
left_leg_EndEffector_torque_X
left_leg_EndEffector_torque_Y
left_leg_EndEffector_torque_Z

devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
HosameldinMohamed: > @traversaro If you call calib all it should not overwrite the offset obtained from the configuration file since it is stored separately. At least that was the intention. This way it kept backward compatibility and the preestimated offset is not lost in case the user inadvertently used calib all.
> 
> @HosameldinMohamed you could verify if its the case.

We couldn't do it this time. We will try to test this feature next time.

fjandrad:
> We expect that the difference between filteredFT data of the external WBD device (pre-estimated offset) and the raw FT data should be the configured value24.8009 , -6.2369 , -62.0044 , -0.0588 , -0.2425 , 0.1253?

Assuming that by raw data you mean one of the analog or measures ports, the difference might not be exactly that because besides the offset there is the secondary calibration matrix. If you want to verify exactly only the offset, then you can put an identity matrix as secondary matrix and a pre-estimated offset or no secondary matrix at all. This would only verify that indeed the pre-estimated offset function is doing what is meant to be.

> And we expect the end effector wrench of the default device to be all of values zero?

If you are checking at the left leg and not the foot then yes it should be zero since there is no force there. Verify that in the settings of the wbd file the frame is lower or upper left leg instead of one of the soles or ankle frames.

> Could it be because we kept the robot running for a long time before dumping the data. So the estimates of both devices drifted a bit?
That is expected since one should be affected by temperature while the other not. The best comparison though is to get something when the robot is cold and recently started and then another one after some time to compare if one stayed around the same values ( the one with temperature compensation ) while the other drifted a bit. It might be worth doing this check when it is on one foot since all the weight will be only on one leg. It should help for repeatability. While in two feet the distribution of forces might be slightly different every time depending on the landing of the robot on the ground.

devices/wholeBodyDynamics/SixAxisForceTorqueMeasureHelpers.cpp
HosameldinMohamed: Yes we noticed that.

HosameldinMohamed:
> Assuming that by raw data you mean one of the analog or measures ports, the difference might not be exactly that because besides the offset there is the secondary calibration matrix. If you want to verify exactly only the offset, then you can put an identity matrix as secondary matrix and a pre-estimated offset or no secondary matrix at all. This would only verify that indeed the pre-estimated offset function is doing what is meant to be.

We commented the configuration groups `FT_SECONDARY_CALIBRATION` and `FT_TEMPERATURE_COEFFICIENTS` to use the default value of `identity` and `0` respectively. In this case we were able to observe the differences between the `filteredFT` port and the `analog:o` port when setting `usePreEstimatedOffset`, and they differences are equal to the values set in the config group `FT_OFFSET`.


Please note that the tests were done using the left-leg FT sensor only.

HosameldinMohamed:
> The best comparison though is to get something when the robot is cold and recently started and then another one after some time to compare if one stayed around the same values ( the one with temperature compensation ) while the other drifted a bit. It might be worth doing this check when it is on one foot since all the weight will be only on one leg.

With @prashanthr05 we recorded the data of the `filteredFT` port of the default `wholeBodyDyanmics` (without the temperature compensation - run in iCub head), and an external `wholeBodyDynamics` with temperature compensation run on an another machine. We recorded the data of both devices when the robot was ***25°C** and when temperature went up to **39°C**. For the device with the temperature compensation we didn't run `usePreEstimatedOffset`.

Here are the data:

### Force X component
| Name | appx. value at 25°C [N] | appx. value at 39°C [N] | difference |
|:--------:|:----------:|:----------:|:-----------:|
| Raw FT data |         12.66         |         13.37         |  -0.71  |
| WBD default |           48.42           |        49.45       |  -1.03 |
| WBD temperature compensation |              24.32               |           23.85             |    0.47   |

 


### Force Y component
| Name | appx. value at 25°C [N] | appx. value at 39°C [N] | difference [N] |
|:--------:|:----------:|:----------:|:-----------:|
| Raw FT data |         26.18         |        22.53         |  3.65   |
| WBD default |           4.25           |        2.434       |  1.816   |
| WBD temperature compensation |              6.5               |           7.85             |  -1.35   |

 

### Force Z component
| Name | appx. value at 25°C [N] | appx. value at 39°C [N] | difference [N] |
|:--------:|:----------:|:----------:|:-----------:|
| Raw FT data |        -213.2         |        -232.5         |   19.3  |
| WBD default |           -285           |        -303.4      |   18.4  |
| WBD temperature compensation |              -210.8               |           -212.4             |  1.6   |

 

### Torque X component
| Name | appx. value at 25°C [N] | appx. value at 39°C [N] | difference [N] |
|:--------:|:----------:|:----------:|:-----------:|
| Raw FT data |        -0.234       |        -0.1236         |   -0.1104  |
| WBD default |           -1.569           |        -1.473      |  -0.096   |
| WBD temperature compensation |              -1.432               |           -1.534             |  0.102   |

 

### Torque Y component
| Name | appx. value at 25°C [N] | appx. value at 39°C [N] | difference [N] |
|:--------:|:----------:|:----------:|:-----------:|
| Raw FT data |        5         |        4.4         |  0.6   |
| WBD default |           4.34           |        3.87      |   0.47 |
| WBD temperature compensation |              5.09               |           4.67             |  0.42   |

 

### Torque Z component
| Name | appx. value at 25°C [N] | appx. value at 39°C [N] | difference [N] |
|:--------:|:----------:|:----------:|:-----------:|
| Raw FT data |        -0.137         |        -0.0128         |   -0.1242  |
| WBD default |           0.0953           |        0.122      |  -0.0267   |
| WBD temperature compensation |              -0.1447              |           -0.0414             |  -0.1033   |

fjandrad:
@HosameldinMohamed @prashanthr05 nice work. The code is working as intended. 
It confirms the behaviour I saw in the past. The z component was always the most affected. I'm glad to see that the secondary calibration matrices also diminish oscillation in the torque values.

traversaro:
> Had the impression the auto generated files where no longer commited. Am I wrong? @traversaro 

It depends on which CMake function are used and with which argument.

prashanthr05:
> > Had the impression the auto generated files where no longer commited. Am I wrong? @traversaro
> 
> It depends on which CMake function are used and with which argument.

@traversaro could you add more details to this? Maybe this could be helpful for all of us in the future dealing with IDL autogenerated files. 

Also, @fjandrad @traversaro if you give a thumbs up on this PR, we could maybe finally merge this.

It must be noted that, 
- the experiments were done only for the left leg FT sensors using the configuration linked in this comment https://github.com/robotology/whole-body-estimators/pull/45#issuecomment-601255577

traversaro:
I will clarify the IDL thing when I will get in front of a real keyboard. In the meanwhile I think it is ok to merge this PR.

fjandrad:
My only concern was the IDL files, so if that is not an issue it is fine for me as well.

prashanthr05:
I rebased over the latest devel and updated changelog. 

We could merge this PR once the CI passes.

prashanthr05:
@fjandrad  @HosameldinMohamed @traversaro Merged into devel, thanks!!

Issue #17, opened 2020-01-29
[wholeBodyDynamics] add the option to change the type of `iDynTree::UnknownWrenchContact`


HosameldinMohamed:
It would be nice if we can assume that only external forces acting on the torso are the thrust forces, for which we know the application point and the force direction, So we can estimate the intensities of more than one wrench per sub-model.

According to the estimation theory behind the algorithm (see the paper [Contact Force Estimations Using Tactile Sensors and Force Torque Sensors](https://www.researchgate.net/profile/Andrea_Del_Prete/publication/236152161_Contact_Force_Estimations_Using_Tactile_Sensors_and_Force_Torque_Sensors/links/00b495166e74a5369d000000.pdf). If the applied external wrenches are assumed to `PURE_FORCE`, the estimator can estimate up to 3 forces per sub-model, and if the applied forces are assumed to be `PURE_FORCE_WITH_KNOWN_DIRECTION`, the estimator can estimate up to 6 forces per sub-model.

The idea is to change the `wholeBodyDynamics` device (Without forcing the software that depends on it to be changed) to have the option to specify the unknown external wrenches from the configuration files.

## Proposed modifications

Here are the proposed modifications to the `wholeBodyDynamics` device:

- The idea is to add some parameters in the configuration file `wholebodydynamics-external.xml` ([example](https://github.com/robotology/codyco-modules/blob/de4546e3f0e4bbf89fc474ccd578febf7836d517/src/devices/wholeBodyDynamics/app/wholebodydynamics-icub-eth-six-fts.xml#L9)) right below the [`defaultContactFrames` parameter](https://github.com/robotology/codyco-modules/blob/de4546e3f0e4bbf89fc474ccd578febf7836d517/src/devices/wholeBodyDynamics/app/wholebodydynamics-icub-eth-six-fts.xml#L9).

The additional parameters could be:

- `overrideContactFrames`:
Similar to `defaultContactFrames` contains vector of strings (name of frames). If this parameter exists, the device will ignore the parameter `defaultContactFrames`.

- `contactWrenchType` could be one of the types:
full = FULL_WRENCH
pure = PURE_FORCE
pureKnown = PURE_FORCE_WITH_KNOWN_DIRECTION

- `contactWrenchDirection`:
[...]

PR #48, opened 2020-03-16
[WBD] Add the option to select the type of the Wrench Contact

HosameldinMohamed:
## Summary

The `wholeBodyDynamics` device assumes that external wrench type is a `full-wrench` (all the components of the forces/torques are unknowns to be estimated). This PR will allow to set a-priori the assumptions about the wrench type (pure forces or pure forces with known direction), from the configuration file. In addition, it will allow to set the position of the applied external wrench. Also, the direction in the case of the pure force with known direction.

The configuration file `wholebodydynamics-external.xml` located in `<installation-directory>/share/yarp/robots/iCubGazeboV2_5/estimator` should be have the following lines added (after the parameter `defaultContactFrames`):
```
<param name="overrideContactFrames">(l_hand,r_hand,chest,l_sole,r_sole,l_upper_leg,r_upper_leg,l_elbow_1,r_elbow_1)</param>
<param name="contactWrenchType">(       pureKnown   ,pureKnown       ,full    ,pureKnown     ,pureKnown     ,pure                 ,pureKnown                ,pureKnown              ,pureKnown)</param>
<param name="contactWrenchDirection">(1,0,0   ,0,5,0    ,0,0,1 ,0,0,1  ,0,0,1  ,0,0,1              ,0,0,1             ,0,0,1           ,0,0,1)</param>
<param name="contactWrenchPosition">(  0,0,0   ,8,1,12    ,10,5,-4 ,8,0,12 ,8,0,12 ,8,0,12             ,0,0,0             ,0,0,0           ,0,0,0)</param>
```

## Details
See https://github.com/robotology/whole-body-estimators/issues/17#issue-558572450.


devices/wholeBodyDynamics/KnownExternalWrench.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+KnownExternalWrench::KnownExternalWrench(std::string parsedframeName, std::string parsedType, std::vector<double> parsedDirection, std::vector<double> parsedPosition)
+{
+    frameName = parsedframeName;
+    type = parsedType;
+    position = parsedPosition;
+    direction = parsedDirection;
+    
+    if(type=="full")
```
prashanthr05: you could add an enum class instead of comparing strings.

devices/wholeBodyDynamics/KnownExternalWrench.h
```diff
@@ -0,0 +1,95 @@
+#ifndef KNOWNEXTERNALWRENCH_H
+#define KNOWNEXTERNALWRENCH_H
+
+#include <array>
+#include <vector>
+
+#include <iDynTree/Estimation/ExtWrenchesAndJointTorquesEstimator.h>
+//#include <iDynTree/Core/Position.h>
+//#include <iDynTree/Core/Direction.h>
```
prashanthr05: These lines can then be removed?

devices/wholeBodyDynamics/KnownExternalWrench.h
```diff
@@ -0,0 +1,95 @@
...
+ * Six Axis FT sensor using an affine function:
+ *
+ * ftNew = secondaryCalibrationMatrix*ftOld - offset
+ *
+ * Where offset is the offset of the F/T sensor (tipically computed
+ * online at given instants where only one external force is known
+ * to be acting on the robot, while secondaryCalibrationMatrix is a
+ * secondary calibration matrix, that is tipically set to the identity.
+ *
+ */
```
prashanthr05: Carried over documentation?

devices/wholeBodyDynamics/KnownExternalWrench.h
```diff
@@ -0,0 +1,95 @@
...
+class KnownExternalWrench
+{
+private:
+    std::string frameName;
+    std::string type;
+    std::vector<double> position;
+    std::vector<double> direction;
```
prashanthr05: Why the choice of using `std::vector` instead of `iDynTree::Position` and `iDynTree::Direction`?

devices/wholeBodyDynamics/KnownExternalWrench.h
```diff
@@ -0,0 +1,95 @@
...
+
+public:
+    uint8_t noOfVariables;
```
prashanthr05: what variables? IMO this variable could be more descriptive.

devices/wholeBodyDynamics/KnownExternalWrench.h
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Default constructor: the secondaryCalibrationMatrix is
+     * set to the identity, while the offset is set to 0.
+     */
```
prashanthr05: Carried over documentation?

devices/wholeBodyDynamics/KnownExternalWrench.h
```diff
@@ -0,0 +1,95 @@
...
+    KnownExternalWrench();
+
+    /**
+     *  Copy constructor
+     * TODO
+     */
+    KnownExternalWrench(std::string parsedframeName, std::string parsedType, std::vector<double> parsedDirection, std::vector<double> parsedPosition);
```
prashanthr05: I wouldn't call this a copy constructor. But I know its the fault of the carried over documentation.

devices/wholeBodyDynamics/KnownExternalWrench.h
prashanthr05: In case, you choose to avoid `iDynTree` objects, then I would suggest using `std::array` instead of `std::vector`. In this way, we make a decision already about the size of the object already during compile-time.

HosameldinMohamed:
CC @gabrielenava

devices/wholeBodyDynamics/KnownExternalWrench.h
HosameldinMohamed: > what variables?

I meant the number of the unknowns in the external wrench (full=6, pureForce=3, pureForceKnownDirection=1).

> IMO this variable could be more descriptive.

We can change it to something like `noOfUnknowns`?

prashanthr05: `nrUnknownsInExtWrench` provides the details we seek I suppose?

devices/wholeBodyDynamics/KnownExternalWrench.h
prashanthr05: Please address or remove all the carried over documentation.

devices/wholeBodyDynamics/KnownExternalWrench.h
HosameldinMohamed: It was a TODO then I forgot about it, I'll update it

devices/wholeBodyDynamics/KnownExternalWrench.cpp
```diff
@@ -0,0 +1,127 @@
...
+    {
+        noOfVariables = 6;
```
prashanthr05: Here lies the confusion, whether it is known variables or unknown variables? and variables in which component? Better to be descriptive than be confused.

devices/wholeBodyDynamics/KnownExternalWrench.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+    else if(type=="pureKnown")
+    {
+        noOfVariables = 1;
+    }
+}
+
+bool KnownExternalWrench::isValid(wholeBodyDynamics::KnownExternalWrench _knownExternalWrench)
+{
+    // For each frame in the paramater overrideContactFrames, there is one string value in the paramater contactWrenchType (one of the 3 types mentioned above), and 3 real numbers in the paramater contactWrenchDirection
```
prashanthr05: paramater ---> parameter

devices/wholeBodyDynamics/KnownExternalWrench.cpp
prashanthr05: Why didn't we choose to use the same enum types as defined in iDynTree? FULL_WRENCH, PURE_FORCE, PURE_FORCE_WITH_KNOWN_DIRECTION ?

devices/wholeBodyDynamics/KnownExternalWrench.cpp
```diff
@@ -0,0 +1,127 @@
...
+    else
+    {
+        return true;
+    }
+}
+
+bool KnownExternalWrench::isDirectionValid()
+{
+    //Make sure that the parsed postion vector is of 3 elements
+    if(direction.size() != 3)
```
prashanthr05: This run-time check can be avoided by using an `std::array`.

devices/wholeBodyDynamics/KnownExternalWrench.cpp
HosameldinMohamed: Correct.
related comment https://github.com/robotology/whole-body-estimators/pull/48#discussion_r393002080

prashanthr05:
The configuration file `wholebodydynamics-external.xml` located in `<installation-directory>/share/yarp/robots/iCubGazeboV2_5/estimator` should be have the following lines added (after the parameter `defaultContactFrames`):
```
<param name="overrideContactFrames">(l_hand,r_hand,chest,l_sole,r_sole,l_upper_leg,r_upper_leg,l_elbow_1,r_elbow_1)</param>
<param name="contactWrenchType">(       pureKnown   ,pureKnown       ,full    ,pureKnown     ,pureKnown     ,pure                 ,pureKnown                ,pureKnown              ,pureKnown)</param>
<param name="contactWrenchDirection">(1,0,0   ,0,5,0    ,0,0,1 ,0,0,1  ,0,0,1  ,0,0,1              ,0,0,1             ,0,0,1           ,0,0,1)</param>
<param name="contactWrenchPosition">(  0,0,0   ,8,1,12    ,10,5,-4 ,8,0,12 ,8,0,12 ,8,0,12             ,0,0,0             ,0,0,0           ,0,0,0)</param>
```

Do you think using a bottle of bottles would offer better separability between two different entities than a single long vector, which is prone to error given there is no proper spacing?

Just a curiosity! if this choice is convenient, we go with it.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,234 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+                return false;
+            }
         }
     }
 
-    // For each submodel, we find the first suitable contact frame
-    // This is a n^2 algorithm, but given that is just used in
-    // configuration it should be ok
-    size_t nrOfSubModels = estimator.submodels().getNrOfSubModels();
+    //A parameter overrideContactFrames exists: ignoring defaultContactFrames and considering defaultContactFrames
```
prashanthr05: consdiering overrideContactFrames

devices/wholeBodyDynamics/UnknownWrenchContact.cpp
```diff
@@ -0,0 +1,64 @@
+#include "UnknownWrenchContact.h"
+
+namespace wholeBodyDynamics
+{
+
+UnknownWrenchContact::UnknownWrenchContact()
```
prashanthr05: I'm now confused. What's the difference between this class and http://wiki.icub.org/codyco/dox/html/idyntree/html/structiDynTree_1_1UnknownWrenchContact.html#a4b96dc6c0dad0f2d4cda49d65381b7d0 ? Why did we choose to define a new data structure for the same entity?

HosameldinMohamed: @fjandrad @prashanthr05 I posted this reply but in the "files changed" page, I am assuming it didn't appear for you guys.

My idea was:
`wholeBodyDynamics::UnknownWrenchContact` inherits from `iDynTree::UnknownWrenchContact` and have in addition other variables and methods, they are:

```cxx
std::string frameName;  
uint8_t nrUnknownsInExtWrench;
constructor(); //it contains different parameters than what iDynTree::UnknownWrenchContact has
isTypeValid();
display();
```

Let me know if you suggest a better way to do it.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
-            yError() << "wholeBodyDynamics : openDefaultContactFrames : missing default contact for submodel composed by the links: ";
-            const iDynTree::Traversal & subModelTraversal = estimator.submodels().getTraversal(subModelIdx);
-            for(iDynTree::TraversalIndex i = 0; i < (iDynTree::TraversalIndex) subModelTraversal.getNrOfVisitedLinks(); i++)
+            // We indicate with FRAME_INVALID_INDEX the fact that we still don't have a override contact for the given submodel
+            subModelIndex2OverrideContact[i].resize(nrOfSubModels,iDynTree::FRAME_INVALID_INDEX);
+
+            size_t subModelIdx = estimator.submodels().getSubModelOfFrame(estimator.model(),overrideContactFramesIdx[i]);
+
+            //check if adding the current contact to this subModel will increase the no of variable to more than 6
+            if(subModelVarSize[subModelIdx]+wrench[i].nrUnknownsInExtWrench > 6)
```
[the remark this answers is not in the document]
HosameldinMohamed: @fjandrad 

> You are enforcing that the number of unkowns remains to be 6, so you can find a unique solution?

Yes, following the recommendation of the [theory paper](https://www.researchgate.net/publication/236152161_Contact_Force_Estimations_Using_Tactile_Sensors_and_Force_Torque_Sensors), see section 3 Discussion and conclusion.




> I think that by default there was a behavior to evenly distribute the forces in case of more than one fully unkown contact. Not ideal, but it allowed having more contacts with some heuristic

I am not aware of that, do you mean when there are 2 full-wrench contacts? and the unknowns count will rise to 12?

The theory paper was talking about evenly distributing the forces. But in the case of 2 pure-force contacts in a submodel. In which case, the number of variables is 6 but the rank of the A matrix drops to 5 (related [comment](https://github.com/dic-iit/element_ironcub-estimation/issues/36#issuecomment-591534146)). So they performed statistical analysis in this particular case if I understood well.

> With your change that behavior would be changed. If we agree this is the new behavior it would make change to add it in some changelog or somewhere so that this is made known to the users.

What is currently implemented before the PR is that (in case of taking the contacts information from the `defaultContactFrame` list), the device assigns only one contact force for each submodel and assumes it's full-wrench. See this part:
https://github.com/robotology/whole-body-estimators/blob/d63c82e8c65d483e218e5c42ecd834bb54612ffd/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L341
And documentation:
https://github.com/robotology/whole-body-estimators/blob/d63c82e8c65d483e218e5c42ecd834bb54612ffd/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h#L131

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
```diff
@@ -707,8 +713,16 @@ class WholeBodyDynamicsDevice :  public yarp::dev::DeviceDriver,
       * specified, and a full wrench on the first suitable frame (i.e. frame belonging to the submodel)
       *  is added.
       */
+     //std::string selectedContactFrames{"default"};
```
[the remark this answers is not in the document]
HosameldinMohamed: done in the next comment.

HosameldinMohamed:
> I saw the changes in general seems alright. I just have doubts about the need to add the frame name. I could not find where do you use it. If its not needed then we could use the iDyntree unkownWrenchContact no?

@fjandrad thanks. In https://github.com/robotology/whole-body-estimators/pull/48#discussion_r395712695 I try to justify why I created another entity.

devices/wholeBodyDynamics/UnknownWrenchContact.cpp
HosameldinMohamed: The variable frameName is used only in `display()`, that I used only for debugging.

The key difference is the constructor of this class. It takes the `type` as string, the position and the direction as vectors of double, and then it creates the `iDynTree` objects (`UnknownWrenchContactType`, Position, and `Direction`). See:
```cxx
UnknownWrenchContact::UnknownWrenchContact(std::string parsedframeName, std::string parsedType, std::vector<double> parsedDirection, std::vector<double> parsedPosition)
```

Also `isTypeValid()`, that checks for the errors.

fjandrad: I see. You probably discussed with @traversaro about it. My question is why couldn't this changes go directly to idyntree? It might be a desirable functionality for people who might use idyntree alone.

prashanthr05:
In the future, I suggest you give more meaningful/descriptive commit messages than

> Address @prashanthr05 comments

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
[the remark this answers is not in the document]
fjandrad: Yes, but from the code point of view I think there is nothing preventing of putting two contacts in the same submodel using the list of default contacts.

devices/wholeBodyDynamics/UnknownWrenchContact.cpp
prashanthr05: True. IMO, it's better to open a PR against iDynTree to update the existing data structure rather than having two of them in two different places with almost the same name and only a minimal updated functionality (most of the checks done in this class can be avoided). That too, given the very minimal changes that this class introduces.

Also here we are introducing some vulnerabilities to cause the error.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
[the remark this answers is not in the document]
fjandrad: Nvm I saw it checks if there is a valid contact already so yes there would only be one contact in case of not using the skin.

devices/wholeBodyDynamics/UnknownWrenchContact.cpp
prashanthr05: When we were first discussing this, I had in mind only filling the iDynTree UnknownWrench object appropriately with the use of existing iDynTree Wrench types. In this way, we reused efficiently the available code and reduced the amount of additional work we had to do.

@HosameldinMohamed  please correct me if I had understood wrongly?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
         {
-            defaultContactFramesIdx.push_back(idx);
+            if( subModelIndex2DefaultContact[subModelIdx] == iDynTree::FRAME_INVALID_INDEX )
+            {
+                yError() << "wholeBodyDynamics : openDefaultContactFrames : missing default contact for submodel composed by the links: ";
+                const iDynTree::Traversal & subModelTraversal = estimator.submodels().getTraversal(subModelIdx);
+                for(iDynTree::TraversalIndex i = 0; i < (iDynTree::TraversalIndex) subModelTraversal.getNrOfVisitedLinks(); i++)
+                {
+                    iDynTree::LinkIndex linkIdx = subModelTraversal.getLink(i)->getIndex();
+                    yError() << "wholeBodyDynamics : openDefaultContactFrames :" << estimator.model().getLinkName(linkIdx);
```
prashanthr05: openDefaultContactFrames ---> openContactFrames

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+        overrideContactFrames.resize(propOverrideContactFrames->size());    
+        for(int ax=0; ax < propOverrideContactFrames->size(); ax++)
+        {
+            overrideContactFrames[ax] = propOverrideContactFrames->get(ax).asString().c_str();
+        }
+        
+        //Check if the parameters `propcontactWrenchType`, `contactWrenchDirection` and `contactWrenchPosition` exist in the configuration file
+        if(propContactWrenchType==0 || propContactWrenchDirection==0 || propContactWrenchPosition==0)
+        {
+            yError() << "wholeBodyDynamics : missing necessary parameters for the overrideContactFrames.";
```
prashanthr05: Missing function name in the error message "openContactFrames "

devices/wholeBodyDynamics/UnknownWrenchContact.cpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+    else
+    {
+        return true;
+    }
+}
+
+void UnknownWrenchContact::display()
+{
+    std::cout << "[KnownExternalWrench] Frame name: " << frameName << std::endl;
```
prashanthr05: KnownExternalWrench ---> UnknownWrenchContact

devices/wholeBodyDynamics/UnknownWrenchContact.cpp
prashanthr05: I think the main purpose of introducing this class, was to check the Yarp configuration parameters which are text-based and fill the corresponding iDynTree object which is enum class-based for wrench types and iDyntree position and direction based. 

I totally see this can be efficiently handled within the openContactFrames(config) function directly, instead of introducing this intermediate class. This way, it could be straight-forward.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+    yarp::os::Bottle *propContactWrenchPosition=prop.find("contactWrenchPosition").asList();
+
+    if(propOverrideContactFrames==0)
     {
-       yError() <<"wholeBodyDynamics : Error parsing parameters: \"defaultContactFrames\" should be followed by a list\n";
-       return false;
+        //overrideContactFrames don't exist -> considering defaultContactFrames
+        if(propDefaultContactFrames==0)
+        {
+            yError() <<"wholeBodyDynamics : Error parsing parameters: \"defaultContactFrames\" should be followed by a list\n";
```
prashanthr05: Please consider adding the function names in these error messages.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+                yWarning() << "Frame " << overrideContactFrames[i] << " not found in the model, discarding it";
+            }
+            else
+            {
+                overrideContactFramesIdx.push_back(idx);
+            }
+        }
+
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// The frames from a single submodel should be no more than 1 if the type is FULL_WRENCH, no more than 3 if the type is PURE_FORCE, and no more than 6 if the type is PURE_FORCE_WITH_KNOWN_DIRECTION.
```
prashanthr05: You could also add the reference paper in this comment.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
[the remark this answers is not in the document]
prashanthr05: > Nvm I saw it checks if there is a valid contact already so yes there would only be one contact in case of not using the skin.

Yes, this is done by https://github.com/robotology/whole-body-estimators/blob/c0ca35e6d7b86a99ed6043aec69462c1cd10f066/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L372

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
```diff
@@ -707,8 +713,16 @@ class WholeBodyDynamicsDevice :  public yarp::dev::DeviceDriver,
...
      std::vector<std::string> defaultContactFrames;
+     std::vector<std::string> overrideContactFrames;
+     std::vector<std::string> contactWrenchType;
+     std::vector<std::vector<double>> contactWrenchDirection;
+     std::vector<std::vector<double>> contactWrenchPosition;
      std::vector<iDynTree::FrameIndex> subModelIndex2DefaultContact;
+     std::vector<std::vector<iDynTree::FrameIndex>> subModelIndex2OverrideContact;
+     bool overrideContactFramesSelected{false};
+     std::vector<int> subModelVarSize;
```
prashanthr05: Reading the source code, I realized that this variable is not descriptive enough.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
```diff
@@ -707,8 +713,16 @@ class WholeBodyDynamicsDevice :  public yarp::dev::DeviceDriver,
...
       *  is added.
       */
+     //std::string selectedContactFrames{"default"};
      std::vector<std::string> defaultContactFrames;
+     std::vector<std::string> overrideContactFrames;
+     std::vector<std::string> contactWrenchType;
+     std::vector<std::vector<double>> contactWrenchDirection;
+     std::vector<std::vector<double>> contactWrenchPosition;
      std::vector<iDynTree::FrameIndex> subModelIndex2DefaultContact;
+     std::vector<std::vector<iDynTree::FrameIndex>> subModelIndex2OverrideContact;
```
prashanthr05: `subModelIndices2OverrideContact` is more descriptive maybe? because we can have multiple contact frames in a submodel depending on the wrench type. Could you also please add a comment here about these variables?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
             {
-                iDynTree::LinkIndex linkIdx = subModelTraversal.getLink(i)->getIndex();
-                yError() << "wholeBodyDynamics : openDefaultContactFrames :" << estimator.model().getLinkName(linkIdx);
+                //subModelIndex2OverrideContact[subModelIdx].push_back(iDynTree::FRAME_INVALID_INDEX);
+                yWarning() << "subModel no " << subModelIdx << " has the maximum number of variables (6). The frame " << overrideContactFrames[i] << " will not be added to it.";
             }
+            else
+            {
+                //add the contact frame
+                //subModelIndex2OverrideContact[subModelIdx].push_back(overrideContactFramesIdx[i]);
```
prashanthr05: the `push_back` will be wrong because the order of the contact frames can be arbitrary and not depend on how the submodels are ordered. Could you please remove this commented line? and add a line of comment here as to why we use array-indexing instead of  push_back?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+        for(size_t subModelIdx = 0; subModelIdx < nrOfSubModels; subModelIdx++)
+        {
+            if(subModelIndex2OverrideContact[subModelIdx].empty() )
+            {
+                yError() << "wholeBodyDynamics : openOverrideContactFrames : missing override contact for submodel composed by the links: ";
+                const iDynTree::Traversal & subModelTraversal = estimator.submodels().getTraversal(subModelIdx);
+                for(iDynTree::TraversalIndex i = 0; i < (iDynTree::TraversalIndex) subModelTraversal.getNrOfVisitedLinks(); i++)
+                {
+                    iDynTree::LinkIndex linkIdx = subModelTraversal.getLink(i)->getIndex();
+                    yError() << "wholeBodyDynamics : openOverrideContactFrames :" << estimator.model().getLinkName(linkIdx);
```
prashanthr05: openOverrideContactFrames ---> openContactFrames

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+            }
+            // We indicate with FRAME_INVALID_INDEX the fact that we still don't have a override contact for the given submodel
+            //subModelIndex2OverrideContact[i].resize(nrOfSubModels,iDynTree::FRAME_INVALID_INDEX);
+        }
 
-            return false;
+        // Let's check that every submodel has an override contact position
+        for(size_t subModelIdx = 0; subModelIdx < nrOfSubModels; subModelIdx++)
+        {
+            if(subModelIndex2OverrideContact[subModelIdx].empty() )
```
prashanthr05: I think it's also safe to check for invalid frame indices.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1418,6 +1601,14 @@ void WholeBodyDynamicsDevice::readContactPoints()
     int numberOfContacts=0;
     size_t nrOfSubModels = estimator.submodels().getNrOfSubModels();
 
+    //construct the wrenches from the global variables
+    wholeBodyDynamics::UnknownWrenchContact wrench[overrideContactFrames.size()];
```
prashanthr05: why do we have to create local variables in multiple functions? Any specific rationale?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1428,17 +1619,40 @@ void WholeBodyDynamicsDevice::readContactPoints()
         {
             if (contactsReadFromSkin.empty())
             {
-
-                for(size_t subModel = 0; subModel < nrOfSubModels; subModel++)
+                if(!overrideContactFramesSelected)
```
prashanthr05: the multiple if...else logic in this function is very confusing and difficult to follow. It becomes even more complex with addition of our new feature. If it's not too much work, could you please add the logic sequence as a short comment at the beginning of this function?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1527,24 +1765,54 @@ void WholeBodyDynamicsDevice::readContactPoints()
...
+                                                                        iDynTree::UnknownWrenchContact(iDynTree::FULL_WRENCH,iDynTree::Position::Zero()));
+                if( !ok )
+                {
+                    yWarning() << "wholeBodyDynamics: Failing in adding default contact for submodel " << subModel;
+                }
             }
+            /*else
+            {
+                yDebug() << "wholeBodyDynamics: number of contacts in submodel "<<subModel<<" = "<<contacts_for_given_subModel[subModel];
+            }*/
```
prashanthr05: maybe this sub-block can be removed?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1428,17 +1619,40 @@ void WholeBodyDynamicsDevice::readContactPoints()
...
+                        }
+                    }
+                }
+                else
+                {
+                    for(size_t subModel = 0; subModel < nrOfSubModels; subModel++)
                     {
-                        yWarning() << "wholeBodyDynamics: Failing in adding default contact for submodel " << subModel;
+                        //iterate for each added contact frame
+                        for(size_t ovFrame = 0; ovFrame < overrideContactFrames.size(); ovFrame++)
```
prashanthr05: can these multiple usages of similar for loops be replaced by a function?

devices/wholeBodyDynamics/UnknownWrenchContact.cpp
HosameldinMohamed: > When we were first discussing this, I had in mind only filling the iDynTree UnknownWrench object appropriately with the use of existing iDynTree Wrench types. In this way, we reused efficiently the available code and reduced the amount of additional work we had to do.
> 
> @HosameldinMohamed please correct me if I had understood wrongly?

@prashanthr05 No that was the idea, but then as you said :point_down: 

> I think the main purpose of introducing this class, was to check the Yarp configuration parameters which are text-based and fill the corresponding iDynTree object which is enum class-based for wrench types and iDyntree position and direction based.

I agree with what you guys suggested to update the class `iDynTree::UnknownWrenchContact` in iDynTree and to add the checks in `openContactFrames(config)` instead of the intermediate class.

prashanthr05:
> In the future, I suggest you give more meaningful/descriptive commit messages than
> 
> > Address @prashanthr05 comments

You could already do so by the following commands,
`git rebase -i HEAD~2` where 2 is the number of previous commits you want to be shown starting from the head.
Then you would have multiple options to either squash commits `s`, reword the commit messages `r`, etc. You replace the tag `pick` against the respective commit with `r` to reword the commit. 

However, by doing so, you are changing history, so you must decide on this carefully.

Alternately you could also use `git amend`

For more info, please see https://help.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message

devices/wholeBodyDynamics/KnownExternalWrench.cpp
HosameldinMohamed: > Why didn't we choose to use the same enum types as defined in iDynTree? FULL_WRENCH, PURE_FORCE, PURE_FORCE_WITH_KNOWN_DIRECTION ?

addressed in https://github.com/robotology/whole-body-estimators/pull/48/commits/edcd509dad868124ec98e627ecd260af425c8ca1


> you could add an enum class instead of comparing strings which might be prone to error.

I am not sure it's worth creating an enum and a map variables just for these 3 words, what do you think?

devices/wholeBodyDynamics/KnownExternalWrench.h
HosameldinMohamed: at the end I settled with `iDynTree` objects (in https://github.com/robotology/whole-body-estimators/pull/48/commits/edcd509dad868124ec98e627ecd260af425c8ca1)

devices/wholeBodyDynamics/UnknownWrenchContact.cpp
HosameldinMohamed: Thanks guys, in https://github.com/robotology/whole-body-estimators/pull/48/commits/edcd509dad868124ec98e627ecd260af425c8ca1, I removed this class and used `iDynTree::UnknownWrenchContact` instead. However, I am not sure it's needed to update `iDynTree::UnknownWrenchContact` anymore.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: unnecessary if...else conditions are removed in https://github.com/robotology/whole-body-estimators/pull/48/commits/edcd509dad868124ec98e627ecd260af425c8ca1

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: removed and global variables are created instead in https://github.com/robotology/whole-body-estimators/pull/48/commits/edcd509dad868124ec98e627ecd260af425c8ca1

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: You're right. done in https://github.com/robotology/whole-body-estimators/pull/48/commits/edcd509dad868124ec98e627ecd260af425c8ca1

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,75 +303,258 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+                subModelIndex2OverrideContact[i][subModelIdx] = overrideContactFramesIdx[i];
+                subModelVarSize[subModelIdx] += wrench[i].nrUnknownsInExtWrench;
+                yInfo() << "Adding frame " << overrideContactFrames[i] << " (idx= " << overrideContactFramesIdx[i] << ") for subModel no " << subModelIdx << " Number of unkowns has become= " << subModelVarSize[subModelIdx];
+            }
+            // We indicate with FRAME_INVALID_INDEX the fact that we still don't have a override contact for the given submodel
+            //subModelIndex2OverrideContact[i].resize(nrOfSubModels,iDynTree::FRAME_INVALID_INDEX);
+        }
 
-            return false;
+        // Let's check that every submodel has an override contact position
```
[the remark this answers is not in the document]
HosameldinMohamed: I agree, in https://github.com/robotology/whole-body-estimators/pull/48/commits/edcd509dad868124ec98e627ecd260af425c8ca1 I removed the duplicate parts.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: Done in https://github.com/robotology/whole-body-estimators/pull/48/commits/38c21cdf2a12d33f0fd9a3f736d69353bb89e86e for the messages in `openContactFrames()`. In general, the the messages in the rest of the device are don't have the name of the function. We may change it in a separate PR if necessary.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: Done in https://github.com/robotology/whole-body-estimators/pull/48/commits/38c21cdf2a12d33f0fd9a3f736d69353bb89e86e.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
HosameldinMohamed: How about `nrUnknownsInSubModel`? Done in https://github.com/robotology/whole-body-estimators/pull/48/commits/38c21cdf2a12d33f0fd9a3f736d69353bb89e86e

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
HosameldinMohamed: in https://github.com/robotology/whole-body-estimators/pull/48/commits/38c21cdf2a12d33f0fd9a3f736d69353bb89e86e I've removed this variable as I used another way. I have created the variable `contactFramesIdxValidForSubModel`

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: right, done in https://github.com/robotology/whole-body-estimators/pull/48/commits/38c21cdf2a12d33f0fd9a3f736d69353bb89e86e

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: After reducing the nested for-loops to one for-loop (in https://github.com/robotology/whole-body-estimators/pull/48/commits/38c21cdf2a12d33f0fd9a3f736d69353bb89e86e), it doesn't feel necessary to put it in a function. what do you think?

HosameldinMohamed:
Hi guys @prashanthr05 @fjandrad , thank you for your comments. In the previous versions I wanted to have minimum modification to the existing code that's why I created unnecessary objects and had so much code duplicated.

In the latest commits (https://github.com/robotology/whole-body-estimators/pull/48/commits/38c21cdf2a12d33f0fd9a3f736d69353bb89e86e), I have removed the class `wholeBodyDynamics::UnknownWrenchContact` and used the members of the struct `iDynTree::UnknownWrenchContact` instead. At this stage. I don't think it's necessary to update `iDynTree::UnknownWrenchContact` to have the members we talked about (`frameName`, `nrOfUnknowns` and `display()`).

The duplication was removed by filling `iDynTree::UnknownWrenchContact` objects for each frame name, whether `defaultContactFrames` or `overrideContactFrames` are used, except for `defaultContactFrames`, we put the type=`iDynTree::FULL_WRENCH`.

I have also modified the publishing function according to https://github.com/robotology/whole-body-estimators/issues/17#issuecomment-612414675 and the preceding discussion. I have published the contact wrenches instead of the net-wrench if `contactsReadFromSkin.empty()` is satisfied. CC @traversaro .

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
```diff
@@ -129,6 +129,10 @@ class wholeBodyDynamicsDeviceFilters
...
+ * | overrideContactFrames |      -   | vector of strings |  -    |    -          | No      | If not data is read from the skin, and this parameter exists, it will override 'defaultContactFrames'  | For each submodel induced by the FT sensor, the frames that belong to that submodel are selected from the list with constraint that the number of variables does not exceed 6. An error is raised if not suitable frame is found for a submodel. |
+ * | contactWrenchType |      -   | vector of strings |  -    |    -          | Yes if 'overrideContactFrames' exists      | It should contain one field for each override frame. The values can be either 'full', 'pure' or 'pureKnown'  | 'full' means the external wrench is full-wrench (6 unknown variables), 'pure' means pure force (3 unknown variables), and 'pureKnown' means pure force with known direction (1 unknown variable). |
+ * | contactWrenchDirection |      -   | vector of doubles |  -    |    -          | Yes if 'overrideContactFrames' exists      | It should contain 3 fields for each override frame.  |     |
```
prashanthr05, line 134: since the size of `contactWrenchDirection` should be the same as the other relevant types, do you think it's valid to say what values that these 3 fields can assume or maybe it can assume any values because it might only be a placeholder for the consistency in the size of the vectors?

prashanthr05:
How do we agree on a valid test or check for merging this PR?

As usual wait for the tests on the robot? @traversaro @fjandrad

traversaro:
> As usual wait for the tests on the robot? @traversaro @fjandrad

In the current situation this is not doable. If it works on iCubGazeboV2_5 with the old configuration file, I think that is ok. If there are regressions on the real robot, we will fix when we will be able to test on a real robot.

prashanthr05:
> In the latest commits (38c21cd), I have removed the class wholeBodyDynamics::UnknownWrenchContact and used the members of the struct iDynTree::UnknownWrenchContact instead. At this stage. I don't think it's necessary to update iDynTree::UnknownWrenchContact to have the members we talked about (frameName, nrOfUnknowns and display()).

Why do you think it's necessary? Technically, yes it wouldn't be necessary, but it would be nice. 
I thought it would lighten the number of additional variables you introduce here in `WholeBodyDynamics.h` and keep it in conatined within the relevant structure.

HosameldinMohamed:
> @HosameldinMohamed Why do you think it's not necessary? Technically, yes it wouldn't be necessary, but it would be nice.
> I thought it would lighten the number of additional variables you introduce here in WholeBodyDynamics.h and keep it in conatined within the relevant structure.

@prashanthr05 You're right. However, when I checked the source code in `iDynTree`, I saw that adding members `nrOfUnknowns` and `frameName` may not be good in terms of consistency of the code.

About the `frameName` variable, the struct [UnknownWrenchContact](https://github.com/robotology/idyntree/blob/83f306bf521a2e89c594017747bda95fe786e346/src/estimation/include/iDynTree/Estimation/ExternalWrenchesEstimation.h#L73) represent `A contact whose wrench is unknown` without associating it to a frame. Then the class [LinkUnknownWrenchContacts](https://github.com/robotology/idyntree/blob/83f306bf521a2e89c594017747bda95fe786e346/src/estimation/include/iDynTree/Estimation/ExternalWrenchesEstimation.h#L128) add the `UnknownWrenchContact`s to the link using [addNewContactInFrame](https://github.com/robotology/idyntree/blob/83f306bf521a2e89c594017747bda95fe786e346/src/estimation/include/iDynTree/Estimation/ExternalWrenchesEstimation.h#L181).

About the `nrOfUnknowns` variable, `iDynTree` computes it at a later stage, using the function [countUnknowns](https://github.com/robotology/idyntree/blob/83f306bf521a2e89c594017747bda95fe786e346/src/estimation/src/ExternalWrenchesEstimation.cpp#L388), which is called when computing the matrices [in this function](https://github.com/robotology/idyntree/blob/83f306bf521a2e89c594017747bda95fe786e346/src/estimation/src/ExternalWrenchesEstimation.cpp#L428).

And for debugging, the class `LinkUnknownWrenchContacts` has a `toString()` method that can be called after adding the contacts.

So I felt the additional members will not be necessary because they are implemented somehow `iDynTree`.

As we saw in our case in `wholeBodyDynamics`, I create the structs `UnknownWrenchContact` in `openContactFrames()` so that I can assign them directly to `addNewContactInFrame` in `readContactPoints()`.

prashanthr05:
> About the frameName variable, the struct UnknownWrenchContact represents A contact whose wrench is unknown without associating it to a frame. Then the class LinkUnknownWrenchContacts add the UnknownWrenchContacts to the link using addNewContactInFrame.

True. However, my argument here is an unknown wrench needs to be identified at some point and needs to be provided with ceratin metadata (for comparisons). An use-case, for example in a contact localization scenario, if we have some contact location semantics (lets now simplify to name), we can just look at the name to match for newly identified unknownwrench and an already identified unknownwrench to see if they associate to the same contact location (more or less analogous to loop detection or place recognition in SLAM).

> About the nrOfUnknowns variable, iDynTree computes it at a later stage, using the function countUnknowns, which is called when computing the matrices in this function.

then, in this case, it would have made more sense to factor out the counting code as a public function within the relevant class so that it can be reused elsewhere. Duplicating code is highly susceptible to errors in the core concepts when the core concepts tend to evolve (in one place and remain outdated in another).

These are my thoughts. 

ALready by removing the extra structs we initially introduce within this PR, the refactored code commits look a lot more cleaner and easy to understand, looking at the gains vs time tradeoffs.

HosameldinMohamed:
> If it works on iCubGazeboV2_5 with the old configuration file, I think that is ok.

I've tested it with a configuration file and it seems ok. I didn't have issues.

> These are my thoughts.
> 
> Already by removing the extra structs we initially introduce within this PR, the refactored code commits look a lot more cleaner and easy to understand, looking at the gains vs time tradeoffs.
> 
> But we can keep these details in mind, if and when, we refactor the software.

They make sense. I would proceed with merging this PR for now. Then we can open issues/PRs for possible enhancements.

Thanks guys.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1531,20 +1719,23 @@ void WholeBodyDynamicsDevice::readContactPoints()
...
+                }
             }
         }
-        /*else
-        {
-             yDebug() << "wholeBodyDynamics: number of contacts in submodel "<<subModel<<" = "<<contacts_for_given_subModel[subModel];
-        }*/
     }
-
+
```
traversaro, line 1738:
```suggestion

```

HosameldinMohamed: Thanks for pointing this out @traversaro, do you have any further comments?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,78 +303,257 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
-        size_t subModelIdx = estimator.submodels().getSubModelOfFrame(estimator.model(),defaultContactFramesIdx[i]);
+        size_t subModelIdx = estimator.submodels().getSubModelOfFrame(estimator.model(),contactFramesIdx[i]);
 
-        // If the subModel of the frame still does not have a default contact, we add it
-        if( subModelIndex2DefaultContact[subModelIdx] == iDynTree::FRAME_INVALID_INDEX )
+        //check if adding the current contact to this subModel will increase the no of variable to more than 6
+        if(nrUnknownsInSubModel[subModelIdx]+nrUnknownsInExtWrench[i] > 6)
         {
-            subModelIndex2DefaultContact[subModelIdx] = defaultContactFramesIdx[i];
+            yWarning() << "subModel no " << subModelIdx << " has the maximum number of variables (6). The frame " << contactFramesNames[i] << " will not be added to it.";
```
traversaro, line 411: What is the rationale in not exiting as early as possible if the configuration file has an error?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,78 +303,257 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+    yarp::os::Bottle *propContactWrenchDirection=prop.find("contactWrenchDirection").asList();
+    yarp::os::Bottle *propContactWrenchPosition=prop.find("contactWrenchPosition").asList();
 
-    defaultContactFrames.resize(propContactFrames->size());
-    for(int ax=0; ax < propContactFrames->size(); ax++)
+    if(propOverrideContactFrames==0)
     {
-        defaultContactFrames[ax] = propContactFrames->get(ax).asString().c_str();
+        //overrideContactFrames don't exist -> considering defaultContactFrames
+        yInfo() << "wholeBodyDynamics : openContactFrames :  Parsing the parameters from defaultContactFrames.\n";
```
traversaro, line 320: This message may be useful for you while developing the feature, but as you may imagine if we print something for every parameter that we parse, the output of the device would became quite confusing.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -303,78 +303,257 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
...
+            yWarning() << "subModel no " << subModelIdx << " has the maximum number of variables (6). The frame " << contactFramesNames[i] << " will not be added to it.";
+        }
+        else
+        {
+            //add the contact frame
+            contactFramesIdxValidForSubModel[i] = contactFramesIdx[i];
+            //add the submodel index to the valid submodel indexes group
+            validSubModelsIndexes[i] = subModelIdx;
+            nrUnknownsInSubModel[subModelIdx] += nrUnknownsInExtWrench[i];
+            yInfo() << "Adding frame " << contactFramesNames[i] << " (idx= " << contactFramesIdx[i] << ") for subModel no " << subModelIdx << " Number of unknowns has become= " << nrUnknownsInSubModel[subModelIdx];
```
traversaro, line 420: I guess this is a debug output that can be removed?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: I understand, we may remove this one and keep the one related to `overrideContactFrames`. What do you think?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: To be honest I haven't thought about it when I put it. I  tried to keep similar logic to what was there before when someone puts an invalid frame name in `defaultContactFrames` (or a frame not existing in the model), the device doesn't exist only throws a warning.
https://github.com/robotology/whole-body-estimators/blob/22302eef59ad2f0b900a662490850206a6aea909/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L332

I think it better to keep it because when I was testing the estimation. I wanted to change one contact from `pureKnown` to `full` to see how the estimation quality will change, and I didn't want to bother checking and removing the rest of the contacts of the same submodel in the list. The device may proceed and remove them for me!

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: I actually found it useful also as a user. It gives you a quick check of the contacts added instead of going through the configuration file to do proof reading.

traversaro: I agree, but then we should print that for all the configuration files.

HosameldinMohamed:
@traversaro 
> Ok for me for merging, except for the part on the publishing external wrenches that is not related to the PR,

I think it's related to the PR because the user can't find this useful if they get `netExternalWrench` of the whole subsystem. We need to publish each external link separately somehow.


> and breaks who reads from that port

Do you think the condition `contactsReadFromSkin.empty()` that I put is enough? Because when it it's `true`, and `defaultsContactsFrames` is chosen, publishing the contact wrenches should be the same as `netExternalWrenches`. What do you think?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
traversaro: I think we should keep consistency with the existing codebase, so do not print anything (as we do for all the other options). However this is a matter of taste, so it is not a big problem even if  the print remains.

traversaro:
> > and breaks who reads from that port
> 
> Do you think the condition `contactsReadFromSkin.empty()` that I put is enough? Because when it it's `true`, and `defaultsContactsFrames` is chosen, publishing the contact wrenches should be the same as `netExternalWrenches`. What do you think?

I do not think that it is not enough to avoid problems in the future, because it is quite a complex, coupled and undocumented behaviour. The semantics of that port in the documentation is clear, and if I read in the docs of the `WBD_OUTPUT_EXTERNAL_WRENCH_PORTS` that it publishes this port publishes the net external wrench applied on a link, it should continue to publish the net external wrench applied on a link. If you want to read as separated contact forces all the contacts that are estimated  on a link, you can use the existing `dynContacts:o` port, or add some new functionality to stream this information in a port (for example for having it in a YARP port that publishes as a  vector. 

To explain why I think the modification in the current form is dangerous, thing about the following: the simulink controllers read the external wrench being applied on  a sole using the `/wholeBodyDynamics/left_leg/cartesianEndEffectorWrench:o` port (relative config: https://github.com/robotology/robots-configuration/blob/v1.15.0/iCubGenova02/estimators/wholebodydynamics.xml#L39 ). As some point in the future, let's say that someone specifies via `overrideContactFrames` that instead of a single wrench (specified via `defaultContactFrames`) the force under the sole should be estimate as two pure forces in  two different points of the sole. Without him suspecting anything, the semantics of the `/wholeBodyDynamics/left_leg/cartesianEndEffectorWrench:o` port will change  from publishing the 6D force/torque vector applied on the sole, to another 6D vector that contains instead two 3D force vectors. The tricky aspect is that the size of the vector will not change, so on the simulink side no error will be raised, but the controller will read completly wrong data.

traversaro:
@HosameldinMohamed we can discuss about this change of behavior of `WBD_OUTPUT_EXTERNAL_WRENCH_PORTS` when you prefer  via Teams, I guess that also @prashanthr05 can join us if he wants.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: removed in https://github.com/robotology/whole-body-estimators/pull/48/commits/a91ed5422a088256a65c7c27d838184f94cb28ba

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: Okay. Let's keep the consistency. I removed them both in https://github.com/robotology/whole-body-estimators/pull/48/commits/a91ed5422a088256a65c7c27d838184f94cb28ba

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: Okay to keep the consistency I removed the line in https://github.com/robotology/whole-body-estimators/pull/48/commits/a91ed5422a088256a65c7c27d838184f94cb28ba

HosameldinMohamed:
> @HosameldinMohamed we can discuss about this change of behavior of WBD_OUTPUT_EXTERNAL_WRENCH_PORTS when you prefer via Teams, I guess that also @prashanthr05 can join us if he wants.

I have implemented one recommendation we agreed on: Added a parameter:

```xml
<!-- "netWrench" or contactWrenches -->
        <param name="outputWrenchPortInfoType">   contactWrenches   </param>
```

It's an optional parameter with a default value `netWrench` which keeps the original behaviour. Selecting `contactWrenches` changes the output of the ports to publish the contact wrenches.
Done in https://github.com/robotology/whole-body-estimators/pull/48/commits/a91ed5422a088256a65c7c27d838184f94cb28ba

---

> If it works on iCubGazeboV2_5 with the old configuration file, I think that is ok.

I have run `yarprobotinterface` with the file installed in `<robotology-superbuild_location>/robotology-superbuild/build/install/share/yarp/robots/iCubGazeboV2_5/`.
Here's a test by applying an [external wrench](https://github.com/robotology/gazebo-yarp-plugins/tree/master/plugins/externalwrench) to the left hand and checking the plot of `/wholeBodyDynamics/left_arm/endEffectorWrench:o`

wbd test PR

---

If you think it's okay we may merge the PR. Thanks.
@prashanthr05 @traversaro @fjandrad

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
```diff
@@ -140,6 +144,7 @@ class wholeBodyDynamicsDeviceFilters
  * |                |   portName_1   | string (name of the port opened to stream the external wrench | - | - | Yes    | Bottle of three elements describing the wrench published on the port: the first element is the link of which the published external wrench is applied. This wrench is expressed around the origin of the frame named as second paramter, and with the orientation of the third parameter.  |  |
  * |                |   ...   | | - | ..                                        | Yes       | ..  |  |
  * |                |   portName_n   | .. | - | -                               | Yes       | ..  | |
+ * | outputWrenchPortInfoType |   -   | string | - | netWrench | No       | Variable for choosing the type of information published in the YARP ports of the group "WBD_OUTPUT_EXTERNAL_WRENCH_PORTS". The user can select between "netWrench" and "contactWrenches"  | If an invalid value is put, the device ignores it and "netWrench" is selected |
```
traversaro, line 147: Can you briefly explain what `netWrench` and `contactWrenches` do?

traversaro:
It seems ok for me!

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
HosameldinMohamed: Done in https://github.com/robotology/whole-body-estimators/pull/48/commits/0f76c5b94fe976b3afc7b5dfd0e3a8cb3adfb022

prashanthr05:
You could update the Changelog and proceed with the merge, thanks! :)

fjandrad:
Should I rebase the temperature PR? @prashanthr05 @traversaro

traversaro:
> 
> 
> Should I rebase the temperature PR? @prashanthr05 @traversaro

I guess it is a good idea to avoid problems during the PR/merge.

PR #52, opened 2020-03-23
Add fail-fast strategy to C++CI workflow

prashanthr05:
This helps avoid canceling other OS builds when one of the builds fail.

See https://github.com/robotology/whole-body-estimators/pull/45#issuecomment-602619220

> > It stopped the others because Windows failed. Is there a way to allow it to check it anyway?
> 
> According to [this answer](https://github.community/t5/GitHub-Actions/Why-a-matrix-step-will-be-canceled-if-another-one-failed/td-p/30920), this is due to a [fail-fast](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) behavior and by default it is set to true. I will open a PR.


prashanthr05:
Not sure, why it added two additional commits? ;D I had done a git pull on devel and branched out of it.

prashanthr05:
The reason why Windows failed  is described in https://github.com/robotology/whole-body-estimators/pull/45#issuecomment-602624564 and relevant links.

Anyways this is a perfect test for the addition of `fail-fast` strategy. Looks like the test passed.

prashanthr05:
I will go ahead and merge this PR. The Windows related vcpkg error will be handled in a new PR.

PR #54, opened 2020-03-25
[Actions] Add cache action to cache source-based dependency builds and add release versioning to source-based dependencies 

prashanthr05:
- Achieve faster workflow CI builds through,
  - pre-compiled vcpkg archives
  - caching source-based dependency builds

- Add release based versioning for source-based dependencies

Inspired and duplicated from
- https://github.com/robotology/idyntree/pull/655
- https://github.com/dic-iit/bipedal-locomotion-controllers/pull/32
- https://github.com/dic-iit/element_jet-cat-turbines/pull/223

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,9 +77,16 @@ jobs:
             sudo apt install git build-essential cmake libace-dev coinor-libipopt-dev  libboost-system-dev libboost-filesystem-dev \
                              libboost-thread-dev liborocos-kdl-dev libeigen3-dev swig qtbase5-dev qtdeclarative5-dev qtmultimedia5-dev libqt5charts5-dev \
                              libxml2-dev liburdfdom-dev libtinyxml-dev liburdfdom-dev liboctave-dev python-dev valgrind libassimp-dev
+                             
+        - name: Cache Source-based dependencies
+          id: cache-source-deps
+          uses: actions/cache@v1
+          with:
+            path: ${{ github.workspace }}/install/deps
+            key: source-deps-${{runner.os}}-vcpkg-robotology-${{env.vcpkg_robotology_TAG}}-ycm-${{env.YCM_TAG}}-yarp-${{env.YARP_TAG}}-icub-main-${{env.ICUB_TAG}}-iDynTree-${{env.iDynTree_TAG}}
                                                         
         - name: Source-based Dependencies [Windows]
-          if: matrix.os == 'windows-latest'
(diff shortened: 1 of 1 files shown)


prashanthr05:
Is there a way to manually re-run a workflow? I can't find any.

prashanthr05:
> Is there a way to manually re-run a workflow? I can't find any.

I would like to test the Cache feature. Is there a clean way to do this? @traversaro

.github/workflows/ci.yml
```diff
@@ -92,72 +117,76 @@ jobs:
...
+            cmake -DCMAKE_PREFIX_PATH=${GITHUB_WORKSPACE}/install/deps \
+                  -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/install/deps ..
             cmake --build . --config ${{matrix.build_type}} --target install
             
             # iDynTree
             cd ${GITHUB_WORKSPACE}
             git clone https://github.com/robotology/iDynTree
             cd iDynTree
+            git checkout ${iDynTree_TAG}
             git checkout devel
```
HosameldinMohamed: Probably you forgot to remove this line.

traversaro:
> 
> 
> > Is there a way to manually re-run a workflow? I can't find any.
> 
> I would like to test the Cache feature. Is there a clean way to do this instead of adding a dummy commit? @traversaro

Not that I am aware of. For testing, I personally  used by fork of the repo,  and then when everything was working I tried in the main repo I was targeting. However, I think that in this case as most of the  code is similar to the one working in iDynTree, we can also think of merging directly.

prashanthr05:
> Not that I am aware of 

A way was shown in https://github.com/robotology/whole-body-estimators/pull/54#discussion_r397845803 ;D

prashanthr05:
Caching was tested.
Build took
 - ubuntu: 1m 42s
 - windows: 4m 56s
 - macOS: 3m 21s

.github/workflows/ci.yml
prashanthr05: I removed this line in the recent commits https://github.com/robotology/whole-body-estimators/commit/0404ba1760e720f3d40a0c74d56f8bce0ec38f5f

traversaro:
> But I fear the cache is using iDynTree devel. Technically, there shouldn't be a problem.

No problem, the cache produced in a PR is ignored in the main branch, so as soon as you merge it a new (correct) cache will be generated, see https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache .

PR #59, opened 2020-04-02
Merge devel into master - addresses important bugfix

prashanthr05:
- Add caching dependencies in build CI/C++
- Fixes the launching of `wholebodydynamics` device in the simulation

Files to be reviewed, 
- CHANGELOG.md


traversaro:
Note that typically there is no need to wait  for reviews for devel --> master merges.

PR #63, opened 2020-04-07
[devices][wholebodydynamics] fix hardcoded low pass filter parameters

### LINKED ISSUE #53: wholebodydynamics critical bug: parameters related to cutoff frequency are  ignored and hardcoded to 3.0 hertz
**TL;DR: Since 2016, despite what it was documented, the `wholebodydynamics` YARP device ignored the parameters `imuFilterCutoffInHz`, `forceTorqueFilterCutoffInHz`, `jointVelFilterCutoffInHz` and `jointAccFilterCutoffInHz`, and so instead used the default value of *3.0 Hertz* (see  https://github.com/robotology/whole-body-estimators/blob/838a66ea9778ee10c291891e6b4c7077a438a052/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L606) for all the low pass filters that filtered the inputs (imu measurements, ft measurements, joint velocity and acceleration measurements) of the torque and external force estimation algorithm.** 

## Detailed Explanation 
These parameters can still be set via the RPC YARP port of wholebodydynamics, but I do not think anyone has been done this in the last few years, so in practice I think this may be affecting any "fast" experiments that used the low-level torque control, for example fast YOGA or torque-based walking. 

## Proposed Solution
Actually implement what is written in the  docs (see https://github.com/robotology/whole-body-estimators/blob/838a66ea9778ee10c291891e6b4c7077a438a052/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h#L125) and make these parameters compulsory, so users can find these parameters in the `wholebodydynamics.xml` file and be immediately aware that filtering is done. Just setting a default value in the code if no parameter is provided is hiding this logic too much. 

## Post mortem 
Why this happened? I checked a bit the original commits, and the commit in which the documentation about the (actually not implemented) parameter was added is https://github.com/robotology/codyco-modules/commit/788b4918125553f8b0b27b2719cd6a53603139da . 

That commit seems to be some kind of compilation hot-fix, in which accidentally I also included the change in the documentation (probably it was a work in progress, and I accidentally added it with `git commit -a` or a careless `git add *`. I think that teaches us a good lesson: no matter **how small a fix seems to be, always do it through a Pull Request**, so at least you have an occasion to self-review it yourself when you merge the PR. In this case, if I had opened a Pull Request, even without an external reviewer I would have probably noticed the wrong documentation I was adding to the repo.

prashanthr05:
Fixes #53

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed launch-wholebodydynamics-`*`.xml configuration files in order to properly open `wholebodydynamics` device without silently skipping it. (See [!56](https://github.com/robotology/whole-body-estimators/pull/56)).
 - Fixed a few YARP related and iDynTree related deprecations in `wholebodydynamics`, `genericSensorClient` and `baseEstimatorV1`. (See [!58](https://github.com/robotology/whole-body-estimators/pull/58)).
+- Fixed a bug in `wholebodydynamics` related to hard-coded cut-off frequency parameters used for low-pass filtering measurements. (See [!63](https://github.com/robotology/whole-body-estimators/pull/63)).
 
 ### Added
 - Github Workflows to check compilation on `ubuntu-latest`, `macOS-latest` and `windows-latest`. (See [!47](https://github.com/robotology/whole-body-estimators/pull/47)).
--- a/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
+++ b/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
@@ -602,11 +602,7 @@ bool WholeBodyDynamicsDevice::loadSettingsFromConfig(os::Searchable& config)
 {
     // Fill setting with their default values
     settings.kinematicSource             = IMU;
-    settings.imuFilterCutoffInHz         = 3.0;
-    settings.forceTorqueFilterCutoffInHz = 3.0;
-    settings.jointVelFilterCutoffInHz    = 3.0;
-    settings.jointAccFilterCutoffInHz    = 3.0;
-
+    
     yarp::os::Property prop;
     prop.fromString(config.toString().c_str());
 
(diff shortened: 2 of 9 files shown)


prashanthr05:
Oh, bug-fixes can be addressed to master right? Should I redirect the PR to master?

traversaro:
> Oh, bug-fixes can be addressed to master right? Should I redirect the PR to master?

As you prefer, especially if we have frequent release it is not a big problem if bug fixes go to `devel` .

traversaro:
> > Oh, bug-fixes can be addressed to master right? Should I redirect the PR to master?
> 
> As you prefer, especially if we have frequent release it is not a big problem if bug fixes go to `devel` .

Even because this may be a bug fix, but it is also a breaking change over existing configuration files stored in robots-configuration https://github.com/robotology/robots-configuration/search?q=IDYNTREE_SKINDYNLIB_LINKS&unscoped_q=IDYNTREE_SKINDYNLIB_LINKS . It would be good if we also updated the configuration files there, or at least open an issue to do so.

prashanthr05:
> Maybe we also update the changelog?

Done in https://github.com/robotology/whole-body-estimators/pull/63/commits/4c68686471643acc5cd0fa3fae2f76f4fc172619.

prashanthr05:
> > > Oh, bug-fixes can be addressed to master right? Should I redirect the PR to master?
> > 
> > 
> > As you prefer, especially if we have frequent release it is not a big problem if bug fixes go to `devel` .
> 
> Even because this may be a bug fix, but it is also a breaking change over existing configuration files stored in robots-configuration https://github.com/robotology/robots-configuration/search?q=IDYNTREE_SKINDYNLIB_LINKS&unscoped_q=IDYNTREE_SKINDYNLIB_LINKS . It would be good if we also updated the configuration files there, or at least open an issue to do so.

I shall open a PR then.

prashanthr05:
> > > > Oh, bug-fixes can be addressed to master right? Should I redirect the PR to master?
> > > 
> > > 
> > > As you prefer, especially if we have frequent release it is not a big problem if bug fixes go to `devel` .
> > 
> > 
> > Even because this may be a bug fix, but it is also a breaking change over existing configuration files stored in robots-configuration https://github.com/robotology/robots-configuration/search?q=IDYNTREE_SKINDYNLIB_LINKS&unscoped_q=IDYNTREE_SKINDYNLIB_LINKS . It would be good if we also updated the configuration files there, or at least open an issue to do so.
> 
> I shall open a PR then.

The relevant changes in configuration files were merged into robotology/robots-configuration.

Shall we merge this PR then?

traversaro:
Ok for me!

prashanthr05:
Merged, thanks @traversaro  @HosameldinMohamed !

PR #68, opened 2020-05-14
Finding Eigen3 and properly link it when using EigenHelpers.

S-Dafarra:
See https://github.com/robotology/osqp-eigen/issues/61 for details. In brief, when using ``OsqpEigen`` in ``devel``, by a chain reaction,  this repo fails to compile.


prashanthr05:
Thank you @S-Dafarra !

PR #71, opened 2020-06-11
Require iDynTree v1.1.0 in ci.yml

prashanthr05:
iDynTree compilation was failing in Windows Visual Studio due to a missing `<string>` header in Triplets.h (Original issue https://github.com/robotology/idyntree/issues/671). This was reflected in our CI workflow as well, See [here](https://github.com/robotology/whole-body-estimators/runs/740774705?check_suite_focus=true). 

This was fixed in https://github.com/robotology/idyntree/pull/672.

Related issue: https://github.com/dic-iit/bipedal-locomotion-framework/pull/49


prashanthr05:
Merged! Thanks for the review @traversaro

Issue #67, opened 2020-04-15
Add a .ini option to skip the initial calibration and just use zero offset


traversaro:
Users of `whole-body-dynamics` tipically execute the `resetOffset` command at the beginning of the simulation, to  set the offset of the FT sensors  in the    `whole-body-dynamics` to zero. This is both boring and error-prone, as any manual step it may be erroneously skipped.  For this reason, it would make sense to have an option in the `.ini` configuration file of `whole-body-dynamics`  to automatically skip the initial calibration in https://github.com/robotology/whole-body-estimators/blob/v0.2.1/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L1181, and instead use the zero offset (as done in https://github.com/robotology/whole-body-estimators/blob/v0.2.1/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L2122 ). 

cc @GiulioRomualdi @Giulero @gabrielenava @HosameldinMohamed @prashanthr05

traversaro:
Clearly then it make sense to have this option enabled for all the "simulated" robots, or at least the one in which the simulated FT sensors do not have any offset to compensate. : )

GiulioRomualdi:
That's really cool! I remember when I was young and naive that I spent two hours before understanding that `resetOffset` was required

prashanthr05:
I have a question regarding this. I am trying to bypass the initial calibration with simulation by turning a flag on/off from the configuration file.

While doing this, I noted that there is a relevant flag `validOffsetAvailable` that needs to be activated for proper completion of the `publishEstimatedQuantities()` method in order to update the output ports. This flag is turned on by the method endCalibration()
https://github.com/robotology/whole-body-estimators/blob/22302eef59ad2f0b900a662490850206a6aea909/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L2281 which is in turn called by `computeCalibration()` method.
https://github.com/robotology/whole-body-estimators/blob/22302eef59ad2f0b900a662490850206a6aea909/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L1884
Does this computeCalibration() method have to be called every iteration ? As done in 
https://github.com/robotology/whole-body-estimators/blob/22302eef59ad2f0b900a662490850206a6aea909/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L1884 

or needs to be called only once?

[...]

PR #72, opened 2020-06-11
Add option to skip initial calibration and use zero offsets

prashanthr05:
This PR addresses https://github.com/robotology/whole-body-estimators/issues/67.

The idl files were autogenerated due to addition of parameter  `skipResetFTSensorOffset` to ` idl/wholeBodyDynamicsSettings/wholeBodyDynamicsSettings.thrift`. The autogenerated files are also committed along with this PR.

Since by default the parameter `skipResetOffset`  is set to `false` (as per the regular workflow), only configuration related to simulation was modified by this PR.   `devices/wholeBodyDynamics/app/wholebodydynamics-icub-external-six-fts-sim.xml`


devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -942,7 +961,7 @@ bool WholeBodyDynamicsDevice::applyLPFSettingsFromConfig(const yarp::os::Propert
     {        
         return false;
     }
-    
+
```
traversaro, line 964:
```suggestion

```

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1363,8 +1382,17 @@ bool WholeBodyDynamicsDevice::attachAll(const PolyDriverList& p)
...
     ok = ok && this->attachAllFTs(p);
     ok = ok && this->attachAllIMUs(p);
-
-    ok = ok && this->setupCalibrationWithExternalWrenchOnOneFrame("base_link",100);
+    
+    if (settings.skipResetFTSensorOffset)
+    {
+        this->setFTSensorOffsetsToZero();
+        validOffsetAvailable = true;
+        ok = ok && true;
```
traversaro, line 1390: It is not really clear what is the  point of this line.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -924,6 +924,25 @@ bool WholeBodyDynamicsDevice::loadSettingsFromConfig(os::Searchable& config)
     if (settings.useJointVelocity) { yInfo() << "wholeBodyDynamics: jointVelFilterCutoffInHz: " <<   settings.jointVelFilterCutoffInHz << " Hz."; }
     if (settings.useJointAcceleration) { yInfo() << "wholeBodyDynamics: jointAccFilterCutoffInHz: " <<   settings.jointAccFilterCutoffInHz << " Hz."; }
     
+    if ( !(prop.check("skipResetOffset") && prop.find("skipResetOffset").isBool()) )
+    {
+        yWarning() << "wholeBodyDynamics: skipResetOffset bool parameter missing.";
+        yWarning() << "wholeBodyDynamics: setting skipResetOffset to the default value of false, please remember to resetOffset through the rpc port.";
```
traversaro, line 930: This is misleading. On the real robot, you don't have to remember to call `resetOffset` .

prashanthr05: Noted. I will remove that line.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
prashanthr05: I added that redundant code to debug and forgot to remove.

traversaro:
Thanks! Minor comments: to be honest, the parameter name  `skipResetFTSensorOffset` probably make sense to existing users that called `resetOffets` for thousands of times in their life, but I think it is misleading for new users, as it does suggest that the "FT Sensor offset reset" is skipped, while this option does exactly the opposite: is uses zero FT sensor offsets (so, "reset" offset) at the device startup, instead of performing calibration. For this reason, I think that calling something like `startWithZeroFTSensorOffsets` or similar (suggestions are welcome) would be more clear. Furthermore, also the description is not really clear: "This allows to skip manual calling of resetOffsets for removeing offsets on FT sensors". I think it  would me more clear to first have a place were we state (also in the general description before this parameter) that at the beginning the device estimates the offset of the FT sensors, unless the option `startWithZeroFTSensorOffsets` is passed. 

Furthermore note that a similar functionality (let user specify user-defined offsets, not just 0) is also provided in https://github.com/robotology/whole-body-estimators/pull/45, so we should consider if  the two options can coexist (probably yes, `startWithZeroFTSensorOffsets` can be an easy option to set all the FT sensor offset to zero (useful in simulation), while the option https://github.com/robotology/whole-body-estimators/pull/45 can be used for more complex settings in which the offset need to be manually inserted. 


@GiulioRomualdi @Giulero @gabrielenava @HosameldinMohamed

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
prashanthr05: Addressed in https://github.com/robotology/whole-body-estimators/pull/72/commits/b495053958d918d01711010f7e32f1ac651fc29c

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
prashanthr05: Addressed in https://github.com/robotology/whole-body-estimators/pull/72/commits/b495053958d918d01711010f7e32f1ac651fc29c

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
prashanthr05: Addressed in https://github.com/robotology/whole-body-estimators/pull/72/commits/b495053958d918d01711010f7e32f1ac651fc29c

prashanthr05:
>  I think that calling something like startWithZeroFTSensorOffsets or similar (suggestions are welcome) would be more clear.

Agreed. I think `startWithZeroFTSensorOffsets` is clear. But I will wait a bit to understand if we get some inputs from other users.

> Furthermore note that a similar functionality (let user specify user-defined offsets, not just 0) is also provided in #45, so we should consider if the two options can coexist (probably yes, startWithZeroFTSensorOffsets can be an easy option to set all the FT sensor offset to zero (useful in simulation), while the option #45 can be used for more complex settings in which the offset need to be manually inserted.

I revisited that PR and noticed that. Indeed, it allows the user to add an `FT_OFFSET` group to specify the custom offsets. I hope we will not be crowding the source code too much by adding a redundant feature (`startWithZeroFTSensorOffsets`).


However, if its agreed to have both the options so that they can be used without any confusion. I will rename the parameter to what is agreed upon. Otherwise, we can close this PR.

traversaro:
> However, if its agreed to have both the options so that they can be used without any confusion, I will rename the parameter to what is agreed upon. 

I agree that having both options make sense, especially for simulation users so that they do not need to add a complex `FT_OFFSET` group when they can just add a simple  option.

traversaro:
cc @lrapetti @diegoferigo

prashanthr05:
> Agreed. I think startWithZeroFTSensorOffsets is clear. But I will wait a bit to understand if we get some inputs from other users.

I will proceed to rename the option with `startWithZeroFTSensorOffsets`.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Added
+- Added 'skipResetOffset' to disable/enable bypassing of resetOffset rpc call for setting FT sensor offsets to zero. (See [!72](https://github.com/robotology/whole-body-estimators/pull/72)).
```
traversaro, line 9:
```suggestion
- Added 'startWithZeroFTSensorOffsets' to disable/enable bypassing of resetOffset rpc call for setting FT sensor offsets to zero. (See [!72](https://github.com/robotology/whole-body-estimators/pull/72)).
```

traversaro:
Approved, but check the CHANGELOG suggestion.

CHANGELOG.md
prashanthr05: Addressed in https://github.com/robotology/whole-body-estimators/pull/72/commits/52813c1dfd0470723a4703d4ff165b81e76bcea6

traversaro:
Perfect!

PR #86, opened 2020-10-12
[workflows] remove use of deprecated set-env in Github Action

prashanthr05:
See,
- https://github.com/robotology/idyntree/pull/748
- https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


traversaro:
I think I did something wrong in the iDynTree PR as it is not working now. however I had followed exactly what was done in https://github.com/robotology/assistive-rehab/commit/c3ad521ed1d61cc6c997200b95b00ee21b3d8c87, so I am not sure what went wrong.

traversaro (approved):
Thanks!

PR #94, opened 2020-11-24
Fix unaddressed yarp::os::ConstString deprecation

prashanthr05:
There were a few unaddressed const string deprecations in virtual analog remapper and clients.


traversaro (approved):

prashanthr05:
Merged, thanks!

PR #109, opened 2021-04-13
[WBD] Add fixed joints l_foot_ft_sensor and r_foot_ft_sensor explicitly

HosameldinMohamed:
Reference issue in `iDynTree` https://github.com/robotology/idyntree/issues/829

The PR appends fixed joints `l_foot_ft_sensor`, `r_foot_ft_sensor` in case they are omitted because the corresponding F/T sensors are not included in the model.

This will enable adding the link frames `l_foot` and `r_foot` in the config group `WBD_OUTPUT_EXTERNAL_WRENCH_PORTS` even when the fee F/T sensors are removed from the URDF model.

### DIFF
--- a/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
+++ b/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
@@ -148,14 +148,17 @@ void addVectorOfStringToProperty(yarp::os::Property& prop, std::string key, std:
     return;
 }
 
-bool getConfigParamsAsList(os::Searchable& config,std::string propertyName , std::vector<std::string> & list)
+bool getConfigParamsAsList(os::Searchable& config,std::string propertyName , std::vector<std::string> & list, bool isRequired)
 {
     yarp::os::Property prop;
     prop.fromString(config.toString().c_str());
     yarp::os::Bottle *propNames=prop.find(propertyName).asList();
     if(propNames==nullptr)
     {
-        yError() <<"WholeBodyDynamicsDevice: Error parsing parameters: \" "<<propertyName<<" \" should be followed by a list\n";
+        if(isRequired)
+        {
+            yError() <<"WholeBodyDynamicsDevice: Error parsing parameters: \" "<<propertyName<<" \" should be followed by a list\n";
+        }
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


traversaro:
I think this change will break the use of wbd for models that do not have the `l_foot_ft_sensor` and `r_foot_ft_sensor` (such as old iCub models or non-iCub models). Rather then hardcoding this list, ideally this should be passed in the conf file as a new parameter `additionalConsideredFixedJoints` or similar.

traversaro (changes requested):
See comment.

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -277,6 +277,11 @@ bool WholeBodyDynamicsDevice::openEstimator(os::Searchable& config)
 
     yInfo() << "wholeBodyDynamics : Loading model from " << modelFileFullPath;
 
+    // Add fixed joints l_foot_ft_sensor, r_foot_ft_sensor in case they are omitted because the corresponding F/T sensors are not included in the model
+    std::vector<std::string> additionalConsideredJoints{"l_foot_ft_sensor", "r_foot_ft_sensor"};
+    // Append additionalConsideredJoints to estimationJointNames
+    estimationJointNames.insert(estimationJointNames.end(), additionalConsideredJoints.begin(), additionalConsideredJoints.end());
```
prashanthr05, line 283: I understand this fix, but I believe this fix is very specific to iCub based URDF models. So maybe we can pass these `additionalConsideredJoints` itself as an optional parameter in the configuration file?

prashanthr05: @traversaro what's your opinion on this?

prashanthr05: Just noticed https://github.com/robotology/whole-body-estimators/pull/109#issuecomment-818592941

traversaro (approved):

prashanthr05 (approved):
Thanks! could you update the Changelog as well?

HosameldinMohamed:
> I think this change will break the use of wbd for models that do not have the `l_foot_ft_sensor` and `r_foot_ft_sensor` (such as old iCub models or non-iCub models). Rather then hardcoding this list, ideally this should be passed in the conf file as a new parameter `additionalConsideredFixedJoints` or similar.

Right, in https://github.com/robotology/whole-body-estimators/pull/109/commits/2c649f903c79fb3afde184f76bf8bc07cf46b9e1 I added it as a parameter. In order to reuse the function `getConfigParamsAsList(getConfigParamsAsList(os::Searchable& config,std::string propertyName , std::vector<std::string> & list)` but avoid the error message https://github.com/robotology/whole-body-estimators/blob/9457244856ae0fba7997228eee51c2d70679c0b0/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L158

In case the parameter is optional (as with`additionalConsideredFixedJoints`), I added an additional boolean function to specify whether the parameter is optional or not.

HosameldinMohamed:
Thanks guys!

PR #115, opened 2021-06-16
Add INTERNAL flag to devices cmake to support proper compilation when some cmake flags are unspecified

prashanthr05:
See https://github.com/robotology/yarp/issues/2606


traversaro (approved):
Probably if we set `INTERNAL` we can remove `DEFAULT ON`?

prashanthr05:
> Probably if we set `INTERNAL` we can remove `DEFAULT ON`?

Technically, yes. Ok, I will remove it.

prashanthr05:
Done in https://github.com/robotology/whole-body-estimators/pull/115/commits/cfa3815f3a8b3d5ef60995ce965130a8a4090bdf

HosameldinMohamed (approved):

PR #117, opened 2021-07-12
Add Conda-based Continuous Integration job

traversaro:
While working on https://github.com/robotology/whole-body-estimators/pull/116, I noticed that the current CI is quite slow. In this PR I added a conda-based CI using the template already followed in other repos that should provide feedback on a PR much faster.

PR #118, opened 2021-07-12
Remove old ci workflow after adding Conda-ci

HosameldinMohamed:
Part of PR #117.

- Updating the changelog with adding the conda-based ci.
- Removing the old CI file to prevent it from running too.


traversaro (approved):

HosameldinMohamed:
I don't find a way to remove an existing [release](https://github.com/robotology/whole-body-estimators/releases/tag/v0.5.1) or modify its associated commit.

Maybe we modify the description of the existing release and leave it as it is?

prashanthr05:
> Maybe we modify the description of the existing release and leave it as it is?

Since, it's only CI related, I think it's ok as it is.

Issue #134, opened 2021-12-24
Compilation error when compiling against YARP 3.6


traversaro:
Error:
~~~
2021-12-24T13:10:43.9642369Z [24/37] Building CXX object devices/wholeBodyDynamics/CMakeFiles/wholeBodyDynamicsDevice.dir/WholeBodyDynamicsDevice.cpp.o
2021-12-24T13:10:43.9644045Z FAILED: devices/wholeBodyDynamics/CMakeFiles/wholeBodyDynamicsDevice.dir/WholeBodyDynamicsDevice.cpp.o 
2021-12-24T13:10:43.9652156Z /usr/bin/c++ -D_USE_MATH_DEFINES -DwholeBodyDynamicsDevice_EXPORTS -I/__w/robotology-superbuild/robotology-superbuild/src/whole-body-estimators/devices/wholeBodyDynamics -I/__w/robotology-superbuild/robotology-superbuild/build/src/whole-body-estimators/idl/wholeBodyDynamicsSettings/autogenerated/include -I/__w/robotology-superbuild/robotology-superbuild/build/src/whole-body-estimators/idl/wholeBodyDynamics_IDLServer/autogenerated/include -I/__w/robotology-superbuild/robotology-superbuild/src/whole-body-estimators/libraries/ctrlLibRT/include -isystem /usr/include/eigen3 -isystem /__w/robotology-superbuild/robotology-superbuild/build/install/include -g -fPIC -std=gnu++1z -MD -MT devices/wholeBodyDynamics/CMakeFiles/wholeBodyDynamicsDevice.dir/WholeBodyDynamicsDevice.cpp.o -MF devices/wholeBodyDynamics/CMakeFiles/wholeBodyDynamicsDevice.dir/WholeBodyDynamicsDevice.cpp.o.d -o devices/wholeBodyDynamics/CMakeFiles/wholeBodyDynamicsDevice.dir/WholeBodyDynamicsDevice.cpp.o -c /__w/robotology-superbuild/robotology-superbuild/src/whole-body-estimators/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
2021-12-24T13:10:43.9659590Z In file included from /__w/robotology-superbuild/robotology-superbuild/src/whole-body-estimators/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp:2:0:
[...]

PR #135, opened 2021-12-24
Fix compilation against YARP 3.6

traversaro:
In YARP 3.6, the Editor internal structure for thrift data structures is now not generated by default. So to fix compilation, we now enable it.

Fix https://github.com/robotology/whole-body-estimators/issues/134 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.6.1] - 2021-12-24
+
+### Fixed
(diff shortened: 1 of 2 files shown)


traversaro:
The CI uses conda so is testing only against YARP 3.5 . Compilation against YARP 3.6 is tested in https://github.com/robotology/robotology-superbuild/actions/runs/1619296009 .

traversaro:
@HosameldinMohamed @prashanthr05 As I guess both of you are busy and we are under christmas, I will proceed with merging and releasing this small fix on my own, sorry for this, but this will permit to unblock the use of the superbuild with YARP 3.6 .

HosameldinMohamed:
Thanks @traversaro! sorry I totally forgot about this one.

PR #113, opened 2021-05-12
[WBD] Add `calibStandingWithJetsiRonCubMk1` rpc command

HosameldinMohamed:
Reference issue https://github.com/dic-iit/element_ironcub-estimation/issues/93

This PR adds another `yarp rpc` fucntion `calibStandingWithJetsiRonCubMk1`, to be used only with `iRonCub-Mk1` model/robot.

- The function is intended to be used when the jets in `iRonCub-Mk1` are ON with idle thrust value.
- The function is similar to `calibStanding` with an addition of other 4 known-wrench contacts.

PR #131, opened 2021-11-23
Switch to generate YARP idl files at every build to fix compatibility with YARP master

traversaro:
While working on https://github.com/robotology/robotology-superbuild/issues/900, I had problems with the idl generated files, as the one generated with YARP 3.5 (I guess) had problems to be compiled with YARP master. To avoid the problem at the root, I switched to generate the idl files at each build (instead of committing them manually) and I removed the autogenerated files that at the moment were committed in the repo. This was done by just generating the idl files in the build directory instead of the source directory.

PR #136, opened 2022-01-24
[WBD] update `calibStandingWithJetsiRonCubMk1` for iRonCub-Mk1_1

HosameldinMohamed:
Related PR #113 which was done for `iRonCub-Mk1`. We modify the `RPC` command to be compatible with `iRonCub-Mk1_1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
-- Updated the `yarp rpc` command `calibStandingWithJetsiRonCubMk1` to be `calibStandingWithJetsiRonCubMk1_1`, in order to perform `calibStanding` for the model `iRonCub-Mk1_1` (See [!136](https://github.com/robotology/whole-body-estimators/pull/136)).
+- Updated the `yarp rpc` command `calibStandingWithJetsiRonCubMk1` to be `calibStandingWithJetsiRonCub`, in order to perform `calibStanding` for the models `iRonCub-Mk1` and `iRonCub-Mk1_1` (See [!136](https://github.com/robotology/whole-body-estimators/pull/136)).
 
 ## [0.6.1] - 2021-01-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


HosameldinMohamed:
It's a simple PR but to be sure I am not forgetting anything!

prashanthr05:
I believe you must also commit the autogenerated IDL files?

HosameldinMohamed:
> I believe you must also commit the autogenerated IDL files?

I suppose not @prashanthr05. After #131 we no longer save the autogenerated files.

traversaro:
Why remove the old method `calibStandingWithJetsiRonCubMk1` ?

HosameldinMohamed:
> Why remove the old method `calibStandingWithJetsiRonCubMk1` ?

I wanted to avoid duplicating the code. Because the only change between the methods is in the idle thrust of the arms jets.

Also, I don't think we will be using `iRonCub-Mk1` again, unless we really have to.
But now that you raised this point. It's better to double-check with the team to have a definitive answer.

traversaro:
> > Why remove the old method `calibStandingWithJetsiRonCubMk1` ?
> 
> I wanted to avoid duplicating the code. Because the only change between the methods is in the idle thrust of the arms jets.

We could have an helper method that just takes idle thrust and arm jets as parameters, and then you call that one on the different `calibtStandingWithJets***` methods?

HosameldinMohamed:
> Are you sure that it is ok to have a breaking change? If you prefer to still have `calibStandingWithJetsiRonCubMk1` you can still keep this method and just call `calibStandingWithJetsiRonCub("mk1")` in its implementation, but this is up do you.

@traversaro I see, Which of the two below did you mean?

- Did you mean the user should be able to call both `calibStandingWithJetsiRonCub` (passing a parameter) and `calibStandingWithJetsiRonCubMk1` (with a deprecation warning maybe?).

- Or did you mean the user should be able to call `calibStandingWithJetsiRonCubMk1` and `calibStandingWithJetsiRonCubMk1_1`, while both internally implement the same method?

Just to understand the better practice. But in this particular case, only the iRonCub team are using it. So I think it should be fine to add a breaking change.

traversaro (approved):

HosameldinMohamed:
Thanks!

traversaro:
Yes I was referring to both cases you were referring to.

HosameldinMohamed:
> Yes I was referring to both cases you were referring to.

Ok. I agree that in general this kind of change can be harmful but in this case it should be ok. I added some documentation of the RPC commands #137.

Issue #141, opened 2022-03-07
Analyze the time profiling of WBD methods 


GiulioRomualdi:
I'm trying to run WBD at 1khz. As discussed in https://github.com/ami-iit/robots-configuration/pull/10 I increased the rate of all the devices used by wbd. 

In the context of xprize I wrote a simple time profiler for wbd https://github.com/robotology/whole-body-estimators/commit/e44b223b3fc1890b8454863d270141c6ed2b5180. Running the profiler with wbd at 1khz I obtain the following result

|         timer    name            |    average time (s)  |   deadline misses      |     last deadline miss time (s)   |
|:------------------------------:|:--------------:|:----------:|:---------------:|
|                       publish| 0.000395487473|         0|              0|
|                compute forces| 0.000179018516|         0|              0|
|                   calibration|    4.18783e-07|         0|              0|
|                    kinematics|  2.4863994e-05|         0|              0|
|                 remove offset| 0.000285472475|         0|              0|
|                contact points|  1.5841566e-05|         0|              0|
|                  read sensors|  2.9965689e-05|         0|              0|
|                           all| 0.000937095005|      8351|    0.001013504|

Accordingly to the table there seems that wbd spends considerably amount of time in publishing and removing offsets (low pass filtering). 
Checking with @prashanthr05 we noticed that a a possible bottleneck is due to yarp-to-idyntree copy

https://github.com/robotology/whole-body-estimators/blob/bf7d5ff892c38ed2346857477b603a3dd386e36e/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L2183-L2292

To avoid these copies we could replace the yarp vectors in ctrLibRT filters with eigen refs/vectors. 

cc @prashanthr05 @HosameldinMohamed @traversaro @isorrentino @S-Dafarra

PR #143, opened 2022-03-08
Attempt to improve the performances of the ctrlLibRT library

GiulioRomualdi:
As a consequence of #141 I tried to improve the performances of the `ctrlLibRT`. Before explaining what I did I would recall that:
⚠️ **we should test this PR on the robot to understand if the performances are improved**

First of all, I tried to remove the copies from yarp to idyntree, thanks to this choice this code
```cpp
iDynTree::toYarp(jointAcc,filters.bufferYarpDofs);
const yarp::sig::Vector & outputJointAcc = filters.jntAccFilter->filt(filters.bufferYarpDofs);
iDynTree::toiDynTree(outputJointAcc,jointAcc);
```
becomes
```cpp
iDynTree::toEigen(jointAcc) = filters.jntAccFilter->filt(iDynTree::toEigen(jointAcc));
```
so I removed two copies every time the joint state is filtered. A similar approach has been used for joint vel, joint acc, accelerometers and gyros, ft.

To achieve that I changed the interface of the `ctrLibRT` to consider `eigen::ref`

### DIFF
--- a/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
+++ b/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
@@ -2203,7 +2203,7 @@ void WholeBodyDynamicsDevice::filterSensorsAndRemoveSensorOffsets()
 
         // Run the filter
         Eigen::Ref<const Eigen::VectorXd> outputFt = filters.forcetorqueFilters[ft]->filt(filters.bufferEigen6);
-        iDynTree::toEigen(filteredFTMeasure.getLinearVec3()) = outputFt.tail<3>();
+        iDynTree::toEigen(filteredFTMeasure.getLinearVec3()) = outputFt.head<3>();
         iDynTree::toEigen(filteredFTMeasure.getAngularVec3()) = outputFt.tail<3>();
 
         filteredSensorMeasurements.setMeasurement(iDynTree::SIX_AXIS_FORCE_TORQUE,ft,filteredFTMeasure);
--- a/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
+++ b/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
@@ -2188,22 +2188,23 @@ void WholeBodyDynamicsDevice::filterSensorsAndRemoveSensorOffsets()
                                   settings.jointAccFilterCutoffInHz);
 
     // Filter and remove offset fromn F/T sensors
+    iDynTree::Wrench rawFTMeasure;
+    iDynTree::Wrench rawFTMeasureWithOffsetRemoved;
+    iDynTree::Wrench filteredFTMeasure;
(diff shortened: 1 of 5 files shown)


devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -2188,22 +2188,23 @@ void WholeBodyDynamicsDevice::filterSensorsAndRemoveSensorOffsets()
...
+        filters.bufferEigen6.tail<3>() = iDynTree::toEigen(rawFTMeasureWithOffsetRemoved.getAngularVec3());
 
         // Run the filter
-        const yarp::sig::Vector & outputFt = filters.forcetorqueFilters[ft]->filt(filters.bufferYarp6);
-
-        iDynTree::Wrench filteredFTMeasure;
-
-        iDynTree::toiDynTree(outputFt,filteredFTMeasure);
+        Eigen::Ref<const Eigen::VectorXd> outputFt = filters.forcetorqueFilters[ft]->filt(filters.bufferEigen6);
+        iDynTree::toEigen(filteredFTMeasure.getLinearVec3()) = outputFt.tail<3>();
```
prashanthr05, line 2206:
```suggestion
        iDynTree::toEigen(filteredFTMeasure.getLinearVec3()) = outputFt.head<3>();
```

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.h
```diff
@@ -112,14 +112,8 @@ class wholeBodyDynamicsDeviceFilters
...
-    ///< Yarp vector buffer of dimension 3
-    yarp::sig::Vector bufferYarp3;
-
-    ///< Yarp vector buffer of dimension 6
-    yarp::sig::Vector bufferYarp6;
-
-    ///< Yarp vector buffer of dimension dofs
-    yarp::sig::Vector bufferYarpDofs;
+    ///< Eigen vector buffer of dimension 6
+    Eigen::Matrix<double, 6, 1> bufferEigen6;
```
prashanthr05, line 116:
```suggestion
    ///< Temporary Eigen vector buffer of dimension 6
    Eigen::Matrix<double, 6, 1> bufferEigen6;
```

traversaro:
>  ⚠️ we should test this PR on the robot to understand if the performances are improved

Based on my experience, the change in performance also on a custom example that uses those classes as the wbd does may be also a good indication in the increase in the performance.

GiulioRomualdi:
I benchmarked the two implementations and I didn't notice any notable timing improvements

traversaro:
Let's see the bright side. If you have a simple executable it is much easier to profile the code with something like callgrind or flamegraphs

traversaro:
Removing myself from the reviewers, feel free to add me back if the PR is being worked on.

traversaro:
@GiulioRomualdi do you think we can close this PR?

traversaro:
> @GiulioRomualdi do you think we can close this PR?

I will take the lack of response for a yes.

PR #144, opened 2022-03-08
Fix Windows CI

traversaro:
`windows-latest` now contains Visual Studio 2022, so to use Visual Studio 2019 let's use `windows-2019`.


prashanthr05 (approved):

HosameldinMohamed (approved):
Thanks @traversaro!, I have been receiving emails about the CI failing lately. I didn't manage to check it!

Issue #158, opened 2022-09-16
Compilation failure against YARP master


traversaro:
See:
~~~
2022-09-16T12:50:48.2730287Z FAILED: devices/virtualAnalogClient/CMakeFiles/virtualAnalogClient.dir/yarp_plugin_virtualAnalogClient.cpp.o 
2022-09-16T12:50:48.2733690Z /usr/share/miniconda3/envs/test/bin/x86_64-conda-linux-gnu-c++ -D_USE_MATH_DEFINES -DvirtualAnalogClient_EXPORTS -I/home/runner/work/robotology-superbuild/robotology-superbuild/src/whole-body-estimators/devices/virtualAnalogClient -isystem /home/runner/work/robotology-superbuild/robotology-superbuild/build/install/include -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /usr/share/miniconda3/envs/test/include -O3 -DNDEBUG -fPIC -std=gnu++17 -MD -MT devices/virtualAnalogClient/CMakeFiles/virtualAnalogClient.dir/yarp_plugin_virtualAnalogClient.cpp.o -MF devices/virtualAnalogClient/CMakeFiles/virtualAnalogClient.dir/yarp_plugin_virtualAnalogClient.cpp.o.d -o devices/virtualAnalogClient/CMakeFiles/virtualAnalogClient.dir/yarp_plugin_virtualAnalogClient.cpp.o -c /home/runner/work/robotology-superbuild/robotology-superbuild/build/src/whole-body-estimators/devices/virtualAnalogClient/yarp_plugin_virtualAnalogClient.cpp
2022-09-16T12:50:48.2736378Z In file included from /home/runner/work/robotology-superbuild/robotology-superbuild/build/install/include/yarp/os/SharedLibraryFactory.h:12,
2022-09-16T12:50:48.2737231Z                  from /home/runner/work/robotology-superbuild/robotology-superbuild/build/install/include/yarp/os/SharedLibraryClassFactory.h:9,
2022-09-16T12:50:48.2738071Z                  from /home/runner/work/robotology-superbuild/robotology-superbuild/build/install/include/yarp/os/SharedLibraryClass.h:10,
[...]

PR #159, opened 2022-09-16
Implement VirtualAnalogClient::getAxes to fix compilation against YARP 3.8

traversaro:
Fix https://github.com/robotology/whole-body-estimators/issues/158 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


HosameldinMohamed (approved):

HosameldinMohamed:
The CI is failing in the build step of the newly added part.

I see it is building against YARP 3.7.. But it's supposed to work with older versions, right?

traversaro:
> I see it is building against YARP 3.7.. But it's supposed to work with older versions, right?

Yes, the CI was failing due a bug on my side, not it should be fixed.

PR #166, opened 2022-11-17
Fix compatibility with YARP 3.8

traversaro:
Fix https://github.com/robotology/whole-body-estimators/issues/165
Fix part of https://github.com/robotology/robotology-superbuild/issues/1303


HosameldinMohamed (approved):
Thanks!

PR #167, opened 2023-02-08
fix attaching both IMU and FTs as MAS network wrappers

HosameldinMohamed:
This is a part of the investigation we are doing with @mebbaid using `ergoCubGazeboV1` https://github.com/icub-tech-iit/ergocub-software/issues/35

We couldn't attach both the IMU and the FTs as MAS.

With the previous logic before this PR, the software doesn't distinguish between them. As it seems both devices types implement all the MAS interfaces, including the interfaces `ISixAxisForceTorqueSensors` and `IThreeAxisLinearAccelerometers`.

### DIFF
--- a/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
+++ b/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
@@ -1,3 +1,4 @@
+#include <yarp/os/Log.h>
 #define SKIN_EVENTS_TIMEOUT 0.2
 #include "WholeBodyDynamicsDevice.h"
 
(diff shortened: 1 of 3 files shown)


devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1740,6 +1744,7 @@ bool WholeBodyDynamicsDevice::attachAllFTs(const PolyDriverList& p)
         return false;
     }
 
+    yDebug() << "Elobaid: remappedMASInterfaces= " << remappedMASInterfaces.ftMultiSensors->getNrOfSixAxisForceTorqueSensors();
```
traversaro, line 1747: Probably this can be deleted?

traversaro (approved):
Just small comment.

traversaro:
For the CI you can see https://github.com/robotology/idyntree-yarp-tools/pull/42 .

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
HosameldinMohamed: Ah yes!, done in https://github.com/robotology/whole-body-estimators/pull/167/commits/fca9e1829ca5c403bd4c87a0a5302466c39c4da7

HosameldinMohamed:
> For the CI you can see [robotology/idyntree-yarp-tools#42](https://github.com/robotology/idyntree-yarp-tools/pull/42) .

Thanks for the info, I'll do it in a new PR to keep it organized!

HosameldinMohamed:
From recent tests with real robots (`ergoCub`, `iCubGenova09`), the issue solved by this PR happens only when attaching network wrappers `multipleAnalogSensorServer` because they implement all the MAS interfaces.

While attaching directly the FT devices `embObjMultipleFTsensors` and the IMU devices `embObjIMU` goes through with no errors because these devices implement only the corresponding interfaces.

CC @GiulioRomualdi

PR #173, opened 2023-09-09
Update WBD launch files

HosameldinMohamed:
This PR updates the launch files for `wholeBodyDynamics` to align them with the latest `robots-configuration` and `icub-models` settings. The modifications include:

- F/T sensors are interfaced as MAS instead of analogSensor devices.
- IMU is interfaced as MAS instead of GenericSensor device.
- IMU frame updated.
- Fixed including the estimator files.

I tested the launch file `launch-wholebodydynamics-icub-six-fts-sim.xml` with `iCubGazebo2_5`, and `launch-wholebodydynamics-icub3-sim.xml` with `iCubGazeboV3`.

For the others, that are supposed to be with real robots, I haven't tested them.

This may fix #169

PR #175, opened 2023-09-12
Fix WBD config files broken in #173

HosameldinMohamed:
I changed the way the config files are called in #173, where I hard-coded the name of the estimator XML file to be called by the launch file. For example `launch-wholebodydynamics-icub-six-fts-sim.xml` calls `wholebodydynamics-icub-external-six-fts-sim.xml`.

But looking at the CMake code, the launch files are installed with the generic name `launch-wholebodydynamics.xml`, and the estimator files are installed as `estimators/wholebodydynamics.xml`, and it's installed for the following real robots `iCubDarmstadt01;iCubGenova02;iCubGenova04;iCubNancy01`, and for the simulated robots: `icubGazeboSim`, `iCubGazeboV2_5`, and `iCubGazeboV3`.

So I reverted to the correct line in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Fix WBD config files broken in #173 (https://github.com/robotology/whole-body-estimators/pull/175).
 
 ## [0.10.0] - 2023-09-11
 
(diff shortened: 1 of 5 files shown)


traversaro (approved):

gabrielenava (approved):
thanks @HosameldinMohamed! I have tested with `iCubGazeboV2_5` and `iCubGazeboV3` and the device now runs also from the home folder.

HosameldinMohamed:
Thanks all, merged..

PR #188, opened 2024-04-18
Add support for nonConsideredAxesPositions parameter

traversaro:


traversaro:
pixi CI has been modified to use https://github.com/robotology/idyntree/pull/1174, so it works. conda CI fails as expected.

traversaro:
conda CI is failing as we do not have an icub-main binary compiled with ipopt 3.14.15 or 3.14.16 . I guess we can just remove icub-main for the time being.

traversaro:
> conda CI is failing as we do not have an icub-main binary compiled with ipopt 3.14.15 or 3.14.16 . I guess we can just remove icub-main for the time being.

Actually we can't, as icub-main is actually a dependency of whole-body-estimator. The pixi CI is working as it is compiling icub-main from source. At this point, I think we can just drop the `conda-ci.yml` as the pixi CI is doing almost the same.

traversaro:
Ok, now the PR is ready for review.

S-Dafarra (approved):

S-Dafarra:
Great!

PR #192, opened 2024-05-02
Fix fatal regression in wholebodydynamics device support of `nonConsideredAxesPositions` parameter

traversaro:
The PR https://github.com/robotology/whole-body-estimators/pull/188 (and hence the v0.11.0 release) contained a critical bug that resulted in all the FT sensors of the specified URDF model to be ignored, resulting in the device not be able to start correctly.

Fix https://github.com/icub-tech-iit/ergocub-software/issues/246 .


traversaro:
The macos fails with:
~~~
Run pixi run download_deps
  × The current platform does not satisfy the minimal virtual package
  │ requirements
  ╰─▶ the project does not support 'osx-arm64'

  × The current platform does not satisfy the minimal virtual package
  │ requirements
  ╰─▶ the project does not support 'osx-arm64'

  × The current platform does not satisfy the minimal virtual package
  │ requirements
  ╰─▶ the project does not support 'osx-arm64'

Error: Process completed with exit code 1.
~~~

This was due to the fact that `macos-latest` switched to `osx-arm64`, that is easy to fix, now the ergocub-software is available on `osx-arm64`  https://github.com/conda-forge/ergocub-software-feedstock/pull/1 .

Issue #32, opened 2018-04-18
Specify used F/T sensors in wholebodydynamics with an explicit parameter


traversaro:
TODO 

cc @gabrielenava @ahmadgazar remember me to update this issue

fjandrad:
For developing the temperature compensation the `multipleAnalogSensorsNames` variable was inserted, it allows to signal which are the sensors that are a MAS. We might include a similar variable for selecting which sensors to use. Things to keep in mind are:
- To keep a similar logic as the other variables we will add the sensors to be considered.
- The new variable should have priority over `multipleAnalogSensorsNames`
- Handle the case where `multipleAnalogSensorsNames` could be a valid name but it was removed by the new variable. Should send a warning or fail if a name is inserted that is not there?

traversaro:
> Should send a warning or fail if a name is inserted that is not there?

Fail. [Fail early, fail fast.](https://en.wikipedia.org/wiki/Fail-fast)

fjandrad:
@prashanthr05 this is the issue about enabling/disabling the ft sensors

traversaro:
@prashanthr05 this is probably related to the problem of excluding the FT sensors on the hip that we were discussing yesterday.

prashanthr05:
> I think this is meant to allow to enable/disable and FT sensor from the perspective of the estimation without the need to modify the urdf model.

Ideally, this would be a nice feature to have. 
But commenting out the sensors in the URDF seems to be a workaround solution for quick hacks.

S-Dafarra:
Let me revive this issue, as it happens often, also during demos, that we need to avoid using an F/T from the robot. Editing the URDF is usually something not so convenient

fjandrad:
Wow its been a while. This takes me back haha. I think I was very close to having this ready or that might just me editing my past memories. What you want to do might be related to this function https://github.com/robotology/whole-body-estimators/blob/5fb9aeeb1cc9282189eee11754c78f685814ed9f/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L767 . I vaguely remember activating one or the other depending if a sensor had the temperature component or not.

fjandrad:
[...]

PR #193, opened 2024-07-05
WholeBodyDynamicsDevice: Remove meaningless check that total number FTs in attached device match the total number of FTs consired by estimation

traversaro:
This is a bug found by @LoreMoretti , basically he was trying to remote just the upper_leg FTs from ergocub from the estimation, but the wbd launch was failing with:

~~~
        yError() << "WholeBodyDynamicsDevice::attachAll Invalid number of MAS FT sensors after remapper";
~~~

However, the check does not make a lot of sense. `nrMASFTSensors` is the total number of FT sensors in all the devices attached to wbd (for ergocub 8), while ` remappedMASInterfaces.ftMultiSensors->getNrOfSixAxisForceTorqueSensors()` and `ftMultipleAnalogSensorNames.size()` are the actual number of sensors considered (removing the `upper_leg` fts as @LoreMoretti was doing, 6).

To be honest, I am wondering how we were removing the FT from the estimation with this bug. Perhaps in all other cases were also either avoiding to pass the upper_leg FTs devices to the attach list, or we removed the FT sensors from the devices opened by the yarprobotinterface?

### DIFF
--- a/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
+++ b/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
@@ -1822,18 +1822,6 @@ bool WholeBodyDynamicsDevice::attachAllFTs(const PolyDriverList& p)
         return false;
     }
 
-    if (nrMASFTSensors != remappedMASInterfaces.ftMultiSensors->getNrOfSixAxisForceTorqueSensors())
-    {
-        yError() << "WholeBodyDynamicsDevice::attachAll Invalid number of MAS FT sensors after remapper";
-        return false;
-    }
-
-    if (nrMASFTSensors != ftMultipleAnalogSensorNames.size())
-    {
-	yError() << "WholeBodyDynamicsDevice::attachAll Invalid number of MAS FT sensor names";
-        return false;
-    }
-
     ftMultipleAnalogSensorIdxMapping.resize(ftMultipleAnalogSensorNames.size());
     for (auto ftDx = 0; ftDx < nrMASFTSensors; ftDx++)
     {


S-Dafarra:
> To be honest, I am wondering how we were removing the FT from the estimation with this bug. Perhaps in all other cases were also either avoiding to pass the upper_leg FTs devices to the attach list, or we removed the FT sensors from the devices opened by the yarprobotinterface?

Both 😅

S-Dafarra:
I was looking at this piece of code
https://github.com/robotology/whole-body-estimators/blob/1fc1d54c9db9183f7a6e4a2f537ea8bef885230c/devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp#L1826-L1836

Maybe it should throw an error if the name is not found?

devices/wholeBodyDynamics/WholeBodyDynamicsDevice.cpp
```diff
@@ -1822,18 +1822,6 @@ bool WholeBodyDynamicsDevice::attachAllFTs(const PolyDriverList& p)
...
-    {
-        yError() << "WholeBodyDynamicsDevice::attachAll Invalid number of MAS FT sensors after remapper";
-        return false;
-    }
-
-    if (nrMASFTSensors != ftMultipleAnalogSensorNames.size())
-    {
-	yError() << "WholeBodyDynamicsDevice::attachAll Invalid number of MAS FT sensor names";
-        return false;
-    }
```
S-Dafarra, line 1835: Should we check that at least ``nrMASFTSensors`` is lower?

traversaro: Until we solve https://github.com/robotology/whole-body-estimators/issues/32 this check is basically https://github.com/robotology/whole-body-estimators/pull/193/files#diff-938e8a6e8fc4690975a0955e2f3a91a8a23c313180dda2dea0fc6cbea41046d2R1806-R1813 .

traversaro:
> Maybe it should throw an error if the name is not found?

I am preparing a different PR to drop the support for reading FT sensor data from IAnalogSensor interfaces, I think it is more appropriate to keep that change there.

S-Dafarra (approved):

traversaro:
> > Maybe it should throw an error if the name is not found?
> 
> I am preparing a different PR to drop the support for reading FT sensor data from IAnalogSensor interfaces, I think it is more appropriate to keep that change there.

Actually that was already removed in https://github.com/robotology/whole-body-estimators/pull/181 . However I indeed preparing a different cleanup PR.

